PR #204, opened 2019-11-14
GStreamer backend improvements

RomanHargrave:
This PR adds the ability to configure the GStreamer output pipeline

It adds a new option to the `[audio]` section of the configuration file, `gst-output`, which allows the user to tweak their audio output pipeline. For example, `rgvolume pre-amp=-2 fallback-gain=-6 ! jackaudiosink client-name=tauon` would apply replaygain in album mode with custom pre-amp and fallback gain and then output audio via JACK.

In addition to the more outward change, this also drastically reduces the GStreamer polling rate which makes volume control seem more responsive.

### DIFF
--- a/t_modules/t_gstreamer.py
+++ b/t_modules/t_gstreamer.py
@@ -51,24 +51,25 @@ def player3(tauon):  # GStreamer
             self.mainloop = GLib.MainLoop()
 
             # Create main "playbin" pipeline thingy for simple playback
-            self.pl = Gst.ElementFactory.make("playbin", "player")
+            self.playbin = Gst.ElementFactory.make("playbin", "player")
+            self._output = Gst.parse_bin_from_description(
+                prefs.gst_output, ghost_unlinked_pads=True)
 
-            # Set up output sink
-            if prefs.gst_sink:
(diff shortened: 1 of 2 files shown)


Taiko2k:
Looks good.

Although one thing missing in regards to how config works, you need to also add a line to *save* the config, the *sync-add* method only loads. See like the line `cf.update_value("use-short-buffering", prefs.short_buffer)` 

This is good as is, though I wonder going forward, might it be preferable to rather than have it like that where the user can specify a custom pipeline, but instead, only expose/abstract those as settings in the UI?

It might be nice to be able to list all outputs, and then set an output. Do you know how we can get a list of outputs/sinks maybe? Like I suppose `pulseaudio, alsa devices, jack`. Then this would remove the new gst-output setting? I'm curious to hear your thoughts.

Also just curious, do you have any reasoning for a default of `fallback-gain=-6`?

RomanHargrave:
I had the fallback-gain set as such partially due to the way my audio is set up (there is no pre-amp after Tauon because I use JACK) - I can easily remove it from the defaults. As for the UI, I think there would be some merit to it - although it might lose some flexibility.

PR #217, opened 2019-12-06
[WIP] french i18n

brunob:
à suivre...

### DIFF
--- a/locale/fr_FR/LC_MESSAGES/tauon.po
+++ b/locale/fr_FR/LC_MESSAGES/tauon.po
@@ -5,8 +5,8 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: \n"
-"POT-Creation-Date: 2020-01-09 18:36+1300\n"
-"PO-Revision-Date: 2020-01-20 10:33+0100\n"
+"POT-Creation-Date: 2020-01-21 17:57+0100\n"
+"PO-Revision-Date: 2020-01-21 18:18+0100\n"
 "Last-Translator: b_b <bruno@eliaz.fr>\n"
 "Language-Team: \n"
 "Language: fr_FR\n"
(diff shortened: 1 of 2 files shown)


Taiko2k:
Unfortunately there's insufficient room for many of these strings. Try compile and test to see where things fit. You can use `msgfmt tauon.po -o tauon.mo` to compile.

You might have to rephrase or change the meaning of some things slightly to fit. Or even mince things grammatically if necessary.

Menus can be widened slightly, particularity the ones that are already small. But for buttons and labels in settings, its harder to make room.

Also you missed a few strings. That's okay, but maybe provide a translation for "Drag and drop to import" since its one of the first things the user sees.

brunob:
Ha, so you mean i have to adapt translation to fit in the same amount of characters there is for English strings ? Not really easy, especially with french, but i can give it a try...

Taiko2k:
Depends on where the string appears.

 - Button labels strictly need same or less characters.
 - Settings box labels and menu text have limited room. 
 - Message box text and tooltips can freely be longer.

brunob:
@Taiko2k maybe you could put metadata in the original po file to inform translators when a string need to be limited ?

Taiko2k:
@brunob Okay, I've figured out how translator comments work now. I'll progressively try add them from now on.

brunob:
@Taiko2k thx for that, i'll finish/adapt the french translation after that.

brunob:
FTR usefull links used for translation : 

* https://github.com/quodlibet/quodlibet/blob/master/quodlibet/po/fr.po
* https://github.com/clementine-player/Clementine/blob/master/src/translations/fr.po
* https://l10n.gnome.org/POT/rhythmbox.master/rhythmbox.master.fr.po

Taiko2k:
Its a little tedious figuring out the range for every string. Easier to just eyeball it. Here are the main places to check how things look:

 - Every page in the **Settings** panel.
 - Rename folder box.
 - Rename tracks box.

I've now got menu label width calculated at runtime so we don't need to worry about them anymore.

___

I've now got the `messages.pot` file located at `locale/messages.pot`. To update the po file, run `msgmerge -U locale/fr_FR/LC_MESSAGES/tauon.po locale/messages.pot`.

And again, to compile `msgfmt locale/fr_FR/LC_MESSAGES/tauon.po -o locale/fr_FR/LC_MESSAGES/tauon.mo`.

brunob:
@Taiko2k nice, i've downloaded the new pot file and merged it with provided command, i can see your comments :) I'll update this as soon as i have a bit of time to complete the "hardest" parts.

Taiko2k:
> Not really easy, especially with french, but i can give it a try...

Maybe I need to consider the possibility of changing the layout of settings to a single column instead of double if this issue is too difficult to work around. Infact I would redesign settings eventually in some way, not sure on how that should look exactly.

brunob:
@Taiko2k no hurry, i've not updated this PR only because of lack of time :)

brunob:
@Taiko2k "et voilà" :) Some items may need verification/amending, but this is a good start.

Taiko2k:
Okay. Not everything is fitting (I had a habit of cramming elements so they just fit without regard to localisation) so I will need to go through and make some adjustments to the layout of things, and I might change some strings also. No big deal as it will make further translations easier if I can fix these issues.

brunob:
Thx for the feedback, i've tried to stay as short as possible, but for some cases it was impossible :\

I have doubts on some items, marked as "not sure" in my .mo file, don't know how i can share these infos.

Taiko2k:
OK I think I've fixed all the layout issues now. Would be nice to redesign the settings box from scratch, but I don't have any design inspiration right now unfortunately.

Except, you have two French translated strings I really need to be shorter, they're `Show playtime lines` and `Show playtime stars`. I don't think it matters too much if shortening makes it a little abstract or ambiguous.

If you could, can you update to the latest commit and merge the latest `messages.pot` file with yours. Then look through again, there's some changes, like the set of strings `Click login to open the last.fm web authorisation` etc is now merged.

> I have doubts on some items, marked as "not sure" in my .mo file

You mean about the source meaning, or translated meaning? It might be OK, I already have a tendency to word the English labels in a weird way.

If its about the source meanings, you could list them here if you want clarification.

___

Btw, thanks for persevering! I think we're almost done. :grimacing:

PR #268, opened 2020-03-28
Swedish translation

eson57:
Hi,
Please add Swedish translation. I like this music player, would be nice to use it in my native language.


Taiko2k:
Nice. Looks good.

Just one thing, is it possible you could reword the translation of `Auto update generated playlists` to something a little shorter? If there's an easy way we could shave `Uppdatera genererade spelningslistor automatiskt` down at least about about 5 characters, that would be convenient.

eson57:
Done! I also found and fixed a small typo.

Taiko2k:
OK, this will be included with the next release, probably within the next few days. When you update at that time, let me know if its working as expected.

PR #296, opened 2020-06-29
Initial adapted translation to pt-BR, added xesam:comment and xesam:genre

msmafra:
- Initial adaptation from the pt-PT/en-US translations to pt-BR - Will revise it in the future;
- Added the xesam:comment and xesam:genre metada so it can be obtained via playerctl, mpris-ctl ro dbus-monitor on Linux;
- Looks like the bus_name object wasn't being used;

Just some to changes to keep using Tauon Music Box with my bash script that puts track information on Obs Studio's text fields and changes the album cover (poor man's plugin).
If it won't be added to some release, how can I compile it? I use the Flatpak version on Fedora Linux.

Thx. Awesome music  player. Side by side with Foobar 2000 on my heart. :)


t_modules/t_dbus.py
```diff
@@ -143,7 +142,10 @@ def update(self, force=False):
                                 'xesam:asText': track.lyrics,
                                 'xesam:autoRating': star_count2(tauon.star_store.get(track.index)),
                                 'xesam:composer': dbus.Array([track.composer]),
-                                'tauon:loved': tauon.love(False, track.index)
+                                'tauon:loved': tauon.love(False, track.index),
+                                # added by msmafra
+                                'xesam:comment': dbus.Array([track.comment]),
+                                'xesam:genre': dbus.Array([track.genre]),
```
Taiko2k, line 149: :heavy_check_mark: This looks fine

t_modules/t_dbus.py
```diff
@@ -187,7 +189,7 @@ def update_loop(self):
                         self.PropertiesChanged('org.mpris.MediaPlayer2.Player', {"LoopStatus": self.get_loop_status()}, [])
 
                     def __init__(self, object_path):
-                        dbus.service.Object.__init__(self, bus, object_path)
+                        dbus.service.Object.__init__(self, bus_name, object_path)
```
Taiko2k: :x: I'm not suure if that is correct, actually, you should probably change it to this:

```python
dbus.service.Object.__init__(self, bus, object_path, bus_name=bus_name)
```

t_modules/t_tagscan.py
```diff
@@ -126,6 +127,7 @@ def __init__(self, file):
 
         self.track_gain = None
         self.album_gain = None
+        self.fullpath = ""
```
Taiko2k: :x: This isn't actually relevant here.

t_modules/t_tagscan.py
```diff
@@ -174,14 +176,14 @@ def read_vorbis(self, f):
...
+                        # self.genre = b.decode("utf-8")
                         genres.append(b.decode())
                     elif a == 'cuesheet':
                         self.cue_sheet = b.decode()
                     elif a == "date":
                         self.date = b.decode("utf-8")
-                    elif a == "comment":
-                        self.comment = b.decode("utf-8")
+                    # elif a == "comment":
+                    #     self.comment = b.decode("utf-8")
```
Taiko2k: :x: It looks like you commented this out and added it again

t_modules/t_tagscan.py
```diff
@@ -214,6 +216,12 @@ def read_vorbis(self, f):
                         self.composer = b.decode("utf-8")
                     elif "fmps_rating" == a:
                         self.misc['FMPS_Rating'] = float(b.decode("utf-8"))
+                    # added by msmafra
+                    elif 'comment' == a:
+                        self.comment = b.decode("utf-8")
+                    elif 'genre' == a:
+                        self.genre = b.decode("utf-8")
```
Taiko2k: :x: Adding ***genre*** read here is redundant, as adding this is handled further down where it is parsed from the genre array. 

(It needs to be done like this because its in specification that there may be **multiple** genre tags in the vorbis comment. So, if we were to directly set the genre here, then only whatever one came last would be the one to have effect.)

Taiko2k:
In regards to Adding pt-BR, just something ill point out, 

Screenshot from 2020-06-30 14-30-15

Here, the width of the labels *samplerate* and *play count* are shaving it a bit wide. If possible, it might look nicer if you could chop those words down somehow. But its acceptable as is. Ideally Tauon would automatically make room, but this is a bit far down the priority list atm.

t_modules/t_dbus.py
msmafra: I am also not sure, but worked that way too. :)

t_modules/t_tagscan.py
msmafra: Leftovers of a test, sorry.

t_modules/t_tagscan.py
msmafra: Removed. :)

t_modules/t_tagscan.py
msmafra: I am not used to git and some times Atom confuses me. I think I meant this:
```
                    elif 'comment' == a:
                        self.comment = b.decode("utf-8")
```
Which I put at the end.
Sorry for the mess. :(

msmafra:
> In regards to Adding pt-BR, just something ill point out,
> 
> Screenshot from 2020-06-30 14-30-15
> 
> Here, the width of the labels _samplerate_ and _play count_ are shaving it a bit wide. If possible, it might look nicer if you could chop those words down somehow. But its acceptable as is. Ideally Tauon would automatically make room, but this is a bit far down the priority list atm.

Changed from Amostragem -> Amostra and from Reproduções -> Reprises

msmafra:
Made the changes. Thanks!

Taiko2k:
There should be no modifications to the file `t_tagscan.py`. I don't see that the changes have any effect and not all changes have been reverted.

PR #299, opened 2020-07-05
Update tauon.po

eson57:
Update Swedish translation from POT.


Taiko2k:
I did a quick look over this translation. Just one small thing I noticed. Could you reduce the length of the string for `"Not enough albums in the playlist!"` by at least 2 characters? It would be okay to replace the meaning with `Not enough albums` or similar if necessary.

PR #318, opened 2020-07-26
03 new fields translated pt_BR, Rsvg warning message

msmafra:
- 03 new fields translated:
    - Folders Remaining, Folder Remaining and Add Playing Spotify Album;
- Added version of Rsvg to stop warning message on t_modules/t_svgout.py.


Taiko2k:
Ive added a few new strings now and updated the pot file. You might want to translate those.

PR #335, opened 2020-08-14
New strings translated in Poedit & avoid Gst warnings

msmafra:
New strings translated in Poedit;
Avoid Gst/GstController warnings at startup;


Taiko2k:
Oh I actually already made that require version change, I just hadn't pushed the change yet. Can you please remove changes to that file. Thanks.

By the way, I have translations on Weblate up and running. https://hosted.weblate.org/engage/tauon-music-box/. So you can use that to do translations. But you can still make PRs here if that's the workflow you prefer.

msmafra:
> Oh I actually already made that require version change, I just hadn't pushed the change yet. Can you please remove changes to that file. Thanks.
Sure! Done! :)
> By the way, I have translations on Weblate up and running. https://hosted.weblate.org/engage/tauon-music-box/. So you can use that to do translations. But you can still make PRs here if that's the workflow you prefer.
I saw one PR related to weblate when I was making mine. I took a like some weeks back to check another OpenSource project and found it confusing. I'll take a look again. :)

Cheers!

Taiko2k:
Actually it would be better if you only use Weblate if you can, this would cause less conflicts.

PR #352, opened 2020-09-19
Sylized the broadcast radio page and optimized it for moble

keystroke3:
Main changes in the files include the moving of everything to the center and placing of the control handles to the right of the cover art. The lyrics were moved below everything and given a fixed height and made to scroll. On mobile everything is stacked vertically and centered (for the most part).
#### To Do
<ul>
<li> Add volume icons next to the slider
</li>
<li> Add fallback cover art</li>
<li> Make optimizations for desktop with small browser window</li>
</ul>


Taiko2k:
Looking good. Here's a few specific issues I see :
 - [Desktop] The right side text+controls group is slightly too low, I think it needs to be a little higher to visually align with the album art.
 - [Desktop] The progress bar height scales with window size but should be fixed in height.
 - [Desktop] The width of the lyrics should probably be wider. Sometimes lines are being split in two when really there's plenty of room. Also, the scroll bars horizontal position slides with the width of the text, but it should probably be fixed. Maybe it would be better to make the max width of the lyrics the width of the album art + controls etc?
 - [Mobile] The lyrics intersect the volume slider.

keystroke3:
I made the corrections you asked for. I made the max width of the desktop lyrics the same as the the content above it. I also made the space between the lyrics and volume slider wider. I hope this fixes some issues.

PR #406, opened 2020-10-08
Maloja Scrobbling

krateng:
This is using the new API URL and JSON

### DIFF
--- a/tauon.py
+++ b/tauon.py
@@ -6716,7 +6716,6 @@ def maloja_scrobble(track):
     d["artist"] = track.artist
     d["title"] = track.title
     d["key"] = prefs.maloja_key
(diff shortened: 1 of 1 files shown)


krateng:
I've also added a [basic UI](https://github.com/krateng/TauonMusicBox/tree/maloja_ui) for it, but obviously it would be understandable if you prefer not to clutter your settings page - let me know if I should add that to the branch as well.

Taiko2k:
> I've also added a basic UI for it

Wow, I'm impressed you figured out how to do that! :exploding_head: Go ahead an add it!

krateng:
Alright, it's added! Hope I didn't break any design paradigm, I just tried to keep it similar to the Spotify one.

PR #424, opened 2020-12-01
Add support for plain password authentication

michirod:
Since Nextcloud airsonic implementation does not support
token-based authentication, this tweak is necessary if you want
to stream from a Nextcloud instance.


almereyda:
The upstream issue for this is https://github.com/owncloud/music/issues/295

PR #450, opened 2021-01-18
Add support for jellyfin music libraries

### LINKED ISSUE #430: [Feature Request] Stream Jellyfin Music
My search on how to use the API for Jellyfin yielded the following results:  
[Reddit Thread about the similarities/differences to Emby API](https://old.reddit.com/r/jellyfin/comments/g3ksve/jellyfin_api_full_documentation/)  
[Python API client mentioned in the above thread](https://github.com/iwalton3/jellyfin-apiclient-python/blob/master/jellyfin_apiclient_python/api.py)  
[Jellyfin terminal Music player mentioned](https://github.com/tryffel/jellycli)  

Much of the API is apparently still the same as Emby, so [this](http://swagger.emby.media/?staticview=true) should apply for the most part. One exception is Authentication, mentioned in the reddit thread above.

ritlew:
I filled in the functions by using the comments provided. Please review closely as I did not thoroughly test it. There are a couple problems:

- Seeking doesn't work very well
- I had to modify the cover art code as I needed to set an auth header to get the album art. I copied what airsonic currently does. Feel free to change that to whatever works

closes #430

### DIFF
--- a/t_modules/t_gstreamer.py
+++ b/t_modules/t_gstreamer.py
@@ -785,7 +785,8 @@ def player3(tauon):  # GStreamer
                     elif self.play_state > 0:
                         if not self.using_cache and pctl.target_object.is_network and \
                                 not pctl.target_object.file_ext == "KOEL" and \
-                                not pctl.target_object.file_ext == "SUB":
+                                not pctl.target_object.file_ext == "SUB" and \
+                                not pctl.target_object.file_ext == "JELY":
 
                             if not os.path.exists(tauon.temp_audio):
                                 os.makedirs(tauon.temp_audio)
--- a/t_modules/t_jellyfin.py
+++ b/t_modules/t_jellyfin.py
@@ -73,21 +73,23 @@ class Jellyfin():
         if not self.connected:
             return ""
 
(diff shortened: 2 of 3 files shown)


Taiko2k:
Seeking is a little bit of an issue. Both backends, GStreamer and Phazor behave a little differently. You should test both.

**Phazor expected behavior:** Seeking should work fine, though it may fail if you try seek right away before its buffered enough.

**GStreamer:** By default it should do a weird thing where it re-buffers the whole track. This is a hack/workaround for servers that don't provide content-length headers. Try test with this disabled and see if that works okay:

Edit `t_modules/t_gstreamer` and duplicate line `788`, replacing `SUB` with `JELY`. Try test seeking with that with a number of tracks and see if it works. If seeking always fails, you can revert this change.

ritlew:
@Taiko2k I see what is happening with the seeking. There is a parameter you can specify to actually retrieve the file instead of transcoding it https://github.com/MediaBrowser/Emby/wiki/Audio-Streaming#seeking. Seeking works perfectly with phazor however with gstreamer "Stall..." is printed in the console endlessly on any attempt to play a song

PR #474, opened 2021-02-25
Fix missing tracks in spotify

kthys:
My proposed fix concerning [this issue](https://github.com/Taiko2k/TauonMusicBox/issues/470)
Tested on Tauon 6.5.3 and is working great

### DIFF
--- a/t_modules/t_spot.py
+++ b/t_modules/t_spot.py
@@ -709,17 +709,19 @@ class SpotCtl:
             nt.artist = track.artists[0].name
         if track.album.artists:
             nt.album_artist = track.album.artists[0].name
+        if track.album.release_date:
             nt.date = track.album.release_date
-            nt.album = track.album.name
-            nt.disc_number = track.disc_number
-            nt.length = track.duration_ms / 1000
-            nt.title = track.name
-            nt.track_number = track.track_number
-          # nt.track_total = total_tracks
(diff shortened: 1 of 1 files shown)


t_modules/t_spot.py
```diff
@@ -693,24 +698,28 @@ def load_track(self, track, update_master_count=True, include_album_url=False):
...
-        nt.track_number = track.track_number
-        # nt.track_total = total_tracks
-        nt.art_url_key = track.album.images[0].url
-        parent = (nt.album_artist + " - " + nt.album).strip("- ")
-        nt.parent_folder_path = parent
-        nt.parent_folder_name = parent
+        if track.artists[0].name:
+            nt.artist = track.artists[0].name
+        if track.album.artists:
+            nt.album_artist = track.album.artists[0].name
```
Taiko2k, line 711: Only this line needs to be indented. The lines below should be indented back.

Taiko2k (changes requested):
Seems alright, just one mistake I can see

kthys:
I have unindented the different variables and fixed two other crashes:

- One due to album images that could be missing in some cases
- Another due to the release date that was missing in some cases, making Tauon crash when ```year_search.search(tr.date)```was called in tauon.py

Taiko2k (approved):

PR #493, opened 2021-04-01
dot desktop files translation

msmafra:
Initial translation of the dot desktop files.


msmafra:
> Seems some garbage has been added to the tauon.py file
I'm not sure what happened. Some Atom crazyness...

Taiko2k (changes requested):
No no this is no good 😁. Looks like the newline at the end of tauon.py has been removed, but that line was intentional.

msmafra:
> No no this is no good . Looks like the newline at the end of tauon.py has been removed, but that line was intentional.

You mean the line 43760? I copied and pasted from the raw file from the repo to avoid any... whatever is happenging.
https://raw.githubusercontent.com/Taiko2k/TauonMusicBox/master/tauon.py

Captura de tela de 2021-04-01 19-44-01

PR #495, opened 2021-04-02
Initial translation for dot desktop files

msmafra:
Initial translation for dot desktop files.


Taiko2k:
Sorry if I'm getting pedantic here, but I'm just thinking you should remove JACK from the keyword lists. Its not really overtly *supported* and I don't wanna give the impression that it is.

PR #511, opened 2021-04-23
Add `album_artist` to scrobble

### LINKED ISSUE #510: Doesn't seem to scrobble last.fm album artist
Please correct me if I'm wrong - but I have some tracks in my library with differing track Artists and Album Artists. This doesn't seem to be scrobbled (on last.fm, the track is linked to an album under the Artist's name, but not the Album Artist's name).

Something like that.

(Fantastic music player by the way, looks really nice!)

AndyThePie:
With apologies to the maintainer - should resolve #510?

> I have some tracks in my library with differing track Artists and Album Artists. This doesn't seem to be scrobbled (on last.fm, the track is linked to an album under the Artist's name, but not the Album Artist's name).

To describe this in more detail: 
image
This track is off of the album "works.thE", with the album artist being "Diverse System". When you click "Go to album" on the scrobble, it takes you here:
image
The scrobble is missing the "Album Artist" property. This fixes that. I think. (It works on my system, at least...)

### DIFF
--- a/tauon.py
+++ b/tauon.py
@@ -6518,7 +6518,7 @@ class LastFMapi:
         try:
             if title != "" and artist != "":
                 if album != "":
-                    if album_artist != artist:
+                    if album_artist and album_artist != artist:
                         self.network.scrobble(artist=artist, title=title, album=album, album_artist=album_artist, timestamp=timestamp)
                     else:
                         self.network.scrobble(artist=artist, title=title, album=album, timestamp=timestamp)
--- a/tauon.py
+++ b/tauon.py
@@ -6510,6 +6510,7 @@ class LastFMapi:
         title = track_object.title
         album = track_object.album
         artist = get_artist_strip_feat(track_object)
+        album_artist = track_object.album_artist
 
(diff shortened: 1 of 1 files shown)


tauon.py
```diff
@@ -6510,14 +6510,18 @@ def scrobble(self, track_object, timestamp=None):
...
+        album_artist = track_object.album_artist
 
         print("submitting scrobble...")
 
         # Act
         try:
             if title != "" and artist != "":
                 if album != "":
-                    self.network.scrobble(artist=artist, title=title, album=album, timestamp=timestamp)
+                    if album_artist != artist:
```
Taiko2k, line 6521: Might be better to change this to:
```python
if album_artist and album_artist != artist:
```
Just in case the API expects a none type instead of empty string

Taiko2k (changes requested):

tauon.py
12beesinatrenchcoat: Yeah, that's reasonable.

Taiko2k (approved):

PR #531, opened 2021-05-17
In progress! Works on Fedora.

RSKYS:
Install script


Taiko2k:
As discussed, I think this is a flawed method that shouldn't be encouraged, at least, not something in the main repo.

PR #553, opened 2021-08-24
scan album artist id3 tag

StrangeSeasOfThought:


Taiko2k:
Oh how could I forget that one. Thanks for the fix!

PR #559, opened 2021-09-05
multiple genres for id3

StrangeSeasOfThought:
Note sure if this is the correct or most elegant way to do this...but now it shows multiple genres again.

### DIFF
--- a/t_modules/t_main.py
+++ b/t_modules/t_main.py
@@ -4329,6 +4329,13 @@ def tag_scan(nt):
                             nt.misc['artists'].extend(frame.text)
                         nt.artist = "; ".join(nt.misc['artists'])
 
+                    frames = tag.getall("TCON")
+                    if frames and len(frames) > 0:
+                        nt.misc['genres'] = []
+                        for frame in frames:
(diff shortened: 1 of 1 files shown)


Taiko2k:
Oh impressive. Looks like you're on the right track there!

Just, can you change it so the array in misc only gets created if there is more than one genre.

Also, elsewhere I used semicolon separation. Like this `"; ".join(genres)`. For consistency we should decide which is better and use one. Semicolon or slash?

StrangeSeasOfThought:
> Just, can you change it so the array in misc only gets created if there is more than one genre.

I tried...even before submitting the pull request. But I always ended up with Tauon displaying only the first genre.

I think slash looks more elegant but yeah, it's probably better to stick with semicolon instead.

Taiko2k:
No I mean you can still temporarily make the array, just don't attach it to .misc if there's not more than 1 item.

In fact I'm pretty sure that misc entry isn't used anywhere, it's just for possible future use.

You can leave or change the separator, up to you. But the above change is requested. (Btw you can add a commit to the same branch and it will update the PR automatically, in case you're not familiar git)

Issue #560, opened 2021-09-07
Queuing an album from the gallery with right-click->add to queue only adds the first track


gigawhitlocks:
Version 6.7.1, using the right-click menu OR dragging an album from the gallery to the queue causes only the first track to be queued.

If you don't have time to work on this, let know if you'd like me to try to fix this myself, as I'd be happy to do so.

Taiko2k:
While this could be a bug, I haven't tested this recently, do you understand how it's intended to work? Only one item gets added but it's in album "mode" as denoted by the coloured be in the right side of the queue. Do you see that bar? (it's green or orange depending on the state)

gigawhitlocks:
Thanks for the reply and apologies in advance if it is indeed user error. When I right-click on the album in gallery view it is added to the queue with an orange bar, and the first track plays, but then it advances to the next album in the queue after the first track is complete

gigawhitlocks:
Alternatively, if I select the album title in the playlist view and drag that to the queue, that is added in album mode correctly, and the whole album plays before the next one in the queue

gigawhitlocks:
![Peek 2021-09-07 14-44](https://user-images.githubusercontent.com/3743903/132402196-5d2e9721-ffa2-41cb-93a0-0a22335d29d1.gif)

Taiko2k:
I see! Odd. I'll look into it when I have time.

Taiko2k:
Okay I had a look and I can replicate it. Bit of a nasty bug since it's not obvious how to replicate it 100% of the time.

I believe I have found the cause and have committed a fix now. From my understanding it only affected the gallery menu entry, but not drag and drop from gallery

gigawhitlocks:
>I believe I have found the cause and have committed a fix now

I just built Tauon from the `master` branch & confirmed that your fix is working on my end as well. Thank you so much, especially for such a quick turnaround! :raised_hands: 

>From my understanding it only affected the gallery menu entry, but not drag and drop from gallery

I believe that's correct, yeah, and in any case if there's a situation where a similar bug is found w/ drag and drop, I'll open a separate issue for it to help you keep things separated

PR #561, opened 2021-09-08
Add missing pygobject dependency to requirements.txt

gigawhitlocks:
While I was testing #560 I noticed that one dependency is missing from `requirements.txt`, so I added it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,5 +11,4 @@ setproctitle
 tekore
 pulsectl
 mutagen
-
-
+PyGObject


Taiko2k:
Hmm, actually I consider pygobject to be a "system dependency". I know its a bit misleading and documentation on why is lacking. Also other scripts rely on this file as well so I can't simply change it.

gigawhitlocks:
It's your project so do what you want but this is the Python library, not the system dependency. I also had to install a system library that correlated to this, but this line will be necessary on all your systems and scripts.

I guarantee if you do `pip freeze` you have it installed

PR #667, opened 2022-02-03
Add install scripts to deploy the repo itself as an desktop app

Zaragata:
I did this for the lazy ppl (myself included) who don't want to use flatpaks + like to build the apps they can't find on their distro.

If you clone the repo with `--depth=1` and just run the script, ta-da: Tauon is deployed on the desktop and you just clean+fetch+pull git to update, then re-run `repo-install.sh` again.

Anyway, sorry if this is more trouble than it's worth... Just though someone else might wanna skip some hours of torture. I mean enjoyable torture; learning is fun!

### DIFF
--- a/extra/repo-install.sh
+++ b/extra/repo-install.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+#
+# A script for the lazy folk and probably a butchering of git.
+#
+# Run this script once you have ALL the dependencies set up.
+#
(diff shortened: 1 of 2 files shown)


Taiko2k:
Seems like that could be useful! Looks good, though maybe could have just created a temporary copy of the sh/.desktop files, modified and used those instead so the repo doesn't need to be modified.

Zaragata:
Ah, yes! That is indeed way cleaner.

I'll get to it right away.

PR #668, opened 2022-02-04
Change `extra/repo-install.sh` to clone `extra/tauonmb.*` intead of changing them directly

Zaragata:
So it leaves the tracked files alone.

I also took the liberty to add the `*.tmp.*` line in `.gitignore`. Is that fine? I'll change it back if is not.

I added it in just in case there is no change in the original `extra/tauonmb.*` files and the user wants to update their repo without worrying about them. Either way, if the user would just run the script again, the `*.tmp.*` files will be overwritten and any problem originated from keeping the temporary files between updates will be gone.

Original idea taken from #667 comment section.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,5 @@ pygettext.py
 lib/
 *.DS_Store
 dist/
+*.tmp.*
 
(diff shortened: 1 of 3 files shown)


Taiko2k:
Might as well just remove those tmp files when done though right?

Zaragata:
Now that you mentioned it, should it remove `user-data` too?

Taiko2k:
> Now that you mentioned it, should it remove `user-data` too?

In this case it will be where the users important app data will be so no.

PR #676, opened 2022-02-08
Instead of deleting `tauonmb.tmp.sh`, track `tauonmb-alt.sh`

Zaragata:
In reality, the parsed script can't be temporary.

#674 made so running `extra/repo-install.sh` deletes all .tmp files after parsing them.
After `tauonmb.tmp.desktop` and the .svgs are installed to `.local/shared`, the desktop launcher will still point to the parsed script.
Without it, Tauon won't open if installed using `repo-install.sh`.

So it's better to have `extra/tauonmb-alt.sh` tracked.

_Addendum: repo-install also didn't overwrite the deployed desktop app, it does now._

### DIFF
--- a/extra/repo-install.sh
+++ b/extra/repo-install.sh
@@ -40,6 +40,7 @@ cp -f extra/tauonmb.{,tmp.}desktop
 sed -i "s+/opt/tauon-music-box/tauonmb+sh $RepoDir/extra/tauonmb-alt+g" extra/tauonmb.tmp.desktop
 
 mkdir -p $ShareDir/{applications,icons/hicolor/scalable/apps}
+[ -e $ShareDir/applications/tauonmb.desktop ] && rm $ShareDir/applications/tauonmb.desktop
 install -Dm755 extra/tauonmb.tmp.desktop $ShareDir/applications/tauonmb.desktop
 install -Dm644 extra/tauonmb{,-symbolic}.svg $ShareDir/icons/hicolor/scalable/apps/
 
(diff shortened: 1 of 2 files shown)


Zaragata:
I just thought about it now:

maybe instead of creating another file to be tracked, just change `extra/tauonmb.sh` so whenever the call fails, it tries to run the repo version. Downside: weird code when deploying to slack or nix.

```
...
python3 /opt/tauon-music-box/tauon.py "$@" || cd $(realpath $(dirname "$0")/..) && python3 tauon.py "$@"
...
```

Or just check if `tauon.py` exists in `/opt/tauon-music-box` and make an if statement... same difference, I guess.

RSKYS:
Hey,
thank you for scripts! they might become useful.
I guess it doesn't need another .sh since they'll get read locally.
Better only add remove before .desktop install for assurance, which didn't have problem in term of overwrite in my own case;
As it's gonna get stopped permanently with `set -e`
I don't think any further changes required for now.
better users stick with flatpak if they use Tauon for serious case!

Zaragata:
Ah, yeah! Didn't see that, my bad

I'll close it :P

RSKYS:
Or maybe mistake I made;
Maybe, maybe cache that remained from official install made conflict at some point, I'm 🚫 t sure.
I'm busy tho, will take a look few days later

Taiko2k:
Hmm. We can just forget about the `.sh` and run the script directly using python `python3 .../tauon.py`. Its not actually needed and will make it tidier.

PR #734, opened 2022-04-12
Parse metadata_block_picture blocks in Ogg

da2x:
The METADATA_BLOCK_PICTURE in an Ogg container is a base64-encoded comment containing a METADATA_BLOCK_PICTURE block identical to FLAC’s format.


da2x:
Wrong.

PR #737, opened 2022-04-13
Replace dot-dot-dot with ellipsis

da2x:
```
sed -i 's/\.\.\./…/g' t_modules/*.py locale/*.pot locale/*/*/*.po
```


Taiko2k:
Doing a blind search and replace is super bad practice IMO. You're basically ganurteeded to introduce hidden bugs.

Also, changing the po/pot IDs is probably bad practice. Those are handled by scripts / the translation app.

Another point, *some* terminals don't like outputting non-ascii characters, so I avoid printing strings with them in them to avoid possible future headaches.

PR #763, opened 2022-05-28
Add requirement to requirements.txt

ACuteWoof:
Installed Tauon Music Box today and I had an error saying that module httpx was not found when I tried to launch it. So I thought I'd add it to the requirements.txt here.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ setproctitle
 pulsectl
 mutagen
 natsort
+httpx


Taiko2k:
It's not a direct dependency, but pip should resolve this automatically.

I assume the issue is you installed from the AUR, which this won't fix, it's an Arch Linux bug. I could add it to the PKGBUILD, I guess I should since it seems like it's not getting offically fixed any time soon.

ACuteWoof:
Oh okay, thank you

PR #766, opened 2022-06-01
Add Windows Support

RSKYS:


compile-phazor.sh
```diff
@@ -3,13 +3,33 @@
...
+        -shared -o libphazor.so -fPIC -Wall -O3 -D AO -g # -Wextra
     mkdir -p lib
     mv libphazor.so lib/libphazor.so
+
+elif [[ "$OSTYPE" == "msys"* ]]; then
+    if [[ "$gcc -v" -eq 0 ]]; then
+        ( PATH="/mingw64/bin:$PATH"
+        gcc src/phazor/kissfft/kiss_fftr.c src/phazor/kissfft/kiss_fft.c src/phazor/phazor.c \
+            `pkg-config --cflags --libs samplerate wavpack libpulse-simple opusfile vorbisfile libmpg123 flac libopenmpt` \
+            -shared -o libphazor.so -fPIC -Wall -O3 -g #-Wextra
```
Taiko2k: Gotta add `-D AO -D WIN` to this line.

RSKYS:
Done.
Make sure you add `mingw-w64-x86_64-pulseaudio` and `gcc` to the MSYS2 doc, the part about pacman installs;
I'll switch to Windows later, but I think this is right order.

RSKYS:
Also `mingw-w64-clang-x86_64-libao` then

RSKYS:
Done.
Make sure you add `gcc` itself to the MSYS2 doc, the part about pacman installs

compile-phazor.sh
```diff
@@ -3,13 +3,37 @@
...
+    gcc src/phazor/kissfft/kiss_fftr.c src/phazor/kissfft/kiss_fft.c src/phazor/phazor.c \
+        `pkg-config --cflags --libs ao wavpack samplerate opusfile vorbisfile libmpg123 flac libopenmpt` \
+        -shared -o libphazor.so -fPIC -Wall -O3 -D AO -g # -Wextra
     mkdir -p lib
     mv libphazor.so lib/libphazor.so
+
+elif [[ "$OSTYPE" == "msys"* ]]; then
+    if [[ "$gcc -v" -eq 0 ]]; then
+	    echo -e " \n== == == == == == == == == == == == == == == == == == == == == == == \n\
+ Make sure you've added '/mingw64/bin' to Windows Environment PATH. \n\
```
Taiko2k: Why is this needed?

RSKYS: When you call gcc by bash via msys, gcc has no idea where compile binaries are;
And later you need the path for Tauon stuffs, so better users add it themselves.

If you don't do this, msys wouldn't call it in PATH so gcc only brings up dependency issues.

PR #818, opened 2022-08-28
Reformat using `black`

dxbednarczyk:
Fixes a discrepancy between files in `t_modules` and `tauon.py` when it comes to quotation mark style.

Also adheres to [significant indentation](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#labels-why-no-backslashes).


Taiko2k:
While I appreciate you might have put time into this, for reasons, as a general rule I don't accept reformat PRs.

PR #923, opened 2023-02-10
Create Sweet.ttheme

the-root-user:
A rather minimalist theme

### DIFF
--- a/theme/Sweet.ttheme
+++ b/theme/Sweet.ttheme
@@ -1,0 +1,121 @@
+# Tauon Music Box theme file
+# by @therootuser_
+
+
+#-Backgrounds-----
+
+27,27,32       player background (top panel)
+28,28,33       playlist panel
+24,24,29       side panel
+26,26,31       bottom panel
+28,28,33       gallery background
+20,20,25       queue panel
+
+#-Frame--------
+
+150,21,21       sep line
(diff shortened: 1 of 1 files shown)


the-root-user:
Preview:
Screenshot-23

Taiko2k:
Looks pretty decent to me. Just two things come to mind:

1. Maybe the track artist should have a colour highlight as well, not just the title?
2. I'm not sure the name is exactly fitting to the theme. How about the name "Ash"?

What are your thoughts?

the-root-user:
Thanks. I don't want the Artist highlight actually. I even wanted to have the listing like "Title Artist" but don't know yet how to do that.
"Ash" is a lovely name.

the-root-user:
Though, (on my side) I changed blur from 9 to 15 and it looks cool (also in above snapshot). With 9, things seemed a bit messy. If you have any plans ... :)

Taiko2k:
> I even wanted to have the listing like "Title Artist" but don't know yet how to do that.

You can sort of do this in the 'columns' view where you can re-order the columns. But the standard view is hard coded.

the-root-user:
You forgot to rename? or decided to keep.
OR should I pull another request

Taiko2k:
Ill do it

PR #931, opened 2023-02-20
Add script for fetching Noto Sans font

dannysu:
The [build instruction for Windows](https://github.com/Taiko2k/TauonMusicBox/wiki/Building-for-Windows) has this manual step. This script automates getting the Noto Sans font from the source.

https://fonts.google.com/noto/specimen/Noto+Sans links to https://github.com/notofonts/noto-source, which then says to go to https://notofonts.github.io/.

### DIFF
--- a/fetch-noto-sans.sh
+++ b/fetch-noto-sans.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+if [ -d fonts ]; then
+  echo "fonts directory already exists"
+  exit 1
+fi
+
+TTF_DIR="fonts/NotoSans/googlefonts/ttf"
+
+mkdir -p fonts
+cd fonts
+git clone --filter=blob:none --no-checkout https://github.com/notofonts/notofonts.github.io.git
+cd notofonts.github.io
+git sparse-checkout set --cone "$TTF_DIR"
+git checkout main
+cp "$TTF_DIR"/*.ttf ..
+cd ..
+rm -rf notofonts.github.io


Taiko2k:
Fonts are a bit of a messy issue. I haven't documented this anywhere but currently only the fonts I included in the exe bundle are the ones that are used.

Screenshot from 2023-03-11 23-16-29

They are:

 -  Noto Sans, (inc Bold, Medium, ExtraCondensed and ExtraCondensedBold)
  - Noto Sans CJK JP versions of the above.

But other language versions won't be used. So importing every Noto font is unnecessary. Also I'm not entirely sure if they can be put in sub folders.

I'm not entirely sure on all this, fontconfig is complicated, and even more complicated on Windows where the behavior is different. But I remember running into performance issues trying to load more fonts.

C0rn3j:
@Taiko2k Did anything change regarding fonts since you wrote this up?

Taiko2k:
@C0rn3j No, still the method I used was to copy the folder from previous installs. I'd suggest we just upload them in an archive somewhere to pull when building.

C0rn3j:
I'm planning to do something similar to this PR but only copy out the relevant files from the repo, when I get to it™

C0rn3j:
image

Are we using the full ones?

C0rn3j:
Superseded by #1361 which downloads the specific supported files in the Windows CI run and adds them as a commented out block in run.sh.

Cloning the massive multiGB repo is a bad idea when we can just grab 20MB of files off GitHub.

PR #942, opened 2023-03-12
Use class instead of tuple for menu items

dannysu:
Instead of using tuples, which is difficult to use once there are many values, use a class instead.
Use `__slots__` for speed and memory savings.


Taiko2k:
Oh wow! Yes this needed fixing. I had written it that way based on my incorrect assumption that using tuples would be more performent, but that turned out to just make it more difficult to maintain. Well done tackling this.

I'll get onto giving this a test and code review soon.

PR #946, opened 2023-03-13
Dev Mode: Default to not saving state

dannysu:
This diff adds code to avoid saving state by default while in dev mode. It can still be enabled via the MENU if needed.

The benefit of this is that when someone is developing or debugging, perhaps repeatedly playing songs, doing that won't artificially change things.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ lib/
 *.DS_Store
 dist/
 *.exe
+.dev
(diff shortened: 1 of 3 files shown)


Taiko2k:
The state will only be read when the app is opened, otherwise saving has no immediate effect. But this could still be useful.

I think saving should default to on, since we're currently detecting "Dev mode" by the presence of a git related file. Not all people who pull from git are developers. Actually it might be better to set this mode another way, perhaps just the presence of a file named "dev" instead of ".gitignore"?

PR #949, opened 2023-03-14
Fix /tmp/TauonMusicBox being used on Windows

dannysu:
#932 fixed the location of the audio-cache dir on Windows, but I found that "/tmp/TauonMusicBox" path is still being used at several places.

Closing the app doesn't properly clean the cache dir after #932.

Fixing this by changing all instances of "/tmp/TauonMusicBox" to be cross-platform compatible.

### DIFF
--- a/t_modules/t_extra.py
+++ b/t_modules/t_extra.py
@@ -32,6 +32,12 @@ import re
 import math
 import threading
 import urllib.parse
+import gi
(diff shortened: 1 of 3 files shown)


dannysu:
I don't have a setup for testing the transcoding code paths, but I confirmed that my cache dir gets properly cleaned up after closing the app with this fix.

Issue #828, opened 2022-09-05
[Feature Request] Use a "proper" build system to simplify dependency handling


jansol:
Packaging Tauon for Nix poses some challenges since Nix aims to lock down things as much as possible. Among other things this means that vendoring dependencies is generally frowned upon and the preferred solution is to have a way of informing the build system of the locations of the (already packaged) dependencies explicitly.

Tauon already makes use of pkg-config for a bunch of dependencies, which is great! However miniaudio and kissfft are currently vendored as git submodules, which makes the Nix build a bit messy due to the submodules affecting the hash of the tauon sources.

In interest of getting rid of the vendoring while avoiding having to write a custom build script that has to be maintained on the packaging side, I'd prefer to have a simple CMakeLists.txt or Makefile instead of the current build-phazor.sh script. This would allow the non pkg-config dependencies to be defined via variables that can then be overridden by us at build time without altering the build files.

Here's an example of what that could roughly look like. Place this next to phazor.c and change the indents to tabs since make requires that but github doesn't let me write it that way:
```Makefile
export MINIAUDIO_INCLUDE_DIR ?= .
export VENDORED_KISSFFT ?= 1
export PREFIX=../..

CFLAGS=-fPIC -Wall -O3 -g #-Wextra

ifeq ($(VENDORED_KISSFFT), 1)
    KISSFFT_DEP = kissfft_static # for depending on the build target
    KISSFFT_FLAGS = -lkissfft_static
else
    KISSFFT_DEP =
    KISSFFT_FLAGS = `pkg-config --cflags --libs -static kissfft`
endif

kissfft_static:
    $(MAKE) -C kissfft 

phazor: ${KISSFFT_DEP}
    cc phazor.c -I${KISSFFT_INCLUDE_DIR} -I${MINIAUDIO_INCLUDE_DIR} -shared -o libphazor.o ${KISSFFT_FLAGS}

install: phazor
    mkdir -p ${PREFIX}/lib
    mv libphazor.so ${PREFIX}/lib/libphazor.so
```

You can then invoke this from as you would the old script with: `make -C src/phazor install` (-C just tells make to switch to the given directory before looking for the Makefile.)

The other common option would be CMake, which is a lot more verbose but also quite a bit easier to understand. Rough example of what that would look like (see the [CMake tutorial](https://cmake.org/cmake/help/latest/guide/tutorial/index.html)):
```cmake
cmake_minimum_required(VERSION 3.10)
project(phazor)

[...]

PR #1086, opened 2024-01-30
Improve support for more easy packaging from Linux distros

stkw0:
Closes: #828

- added a CMakeLists.txt to build phrazor component. 
- added a setup.py to build the t_module python module to be able to be installed with the standard approaches of different Linux distros.
- modified install_directory when the program is installed under /usr so it seeks assets in the standard /usr/share directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,19 @@
+cmake_minimum_required(VERSION 3.10)
+project(phazor)
+
+add_library(phazor SHARED src/phazor/phazor.c)
+
+find_package(PkgConfig REQUIRED)
+
+pkg_check_modules(phazor REQUIRED flac libopenmpt libmpg123 opus opusfile wavpack samplerate)
+
+find_package(kissfft CONFIG COMPONENTS float)
+if(NOT kissfft_FOUND)
+	add_subdirectory(src/phazor/kissfft)
+endif()
+
+target_include_directories(phazor PRIVATE ${phazor_INCLUDE_DIRS})
+target_link_libraries(phazor kissfft::kissfft ${phazor_LIBRARIES})
+
+
+install(TARGETS phazor DESTINATION ${CMAKE_SOURCE_DIR}/lib)
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,16 @@
+from setuptools import setup, find_packages, Extension
+
+with open('requirements.txt') as f:
+    required = f.read().splitlines()
+
+setup(
+    name = 'Tauon Music Box',
+    packages = find_packages(),
+    version = '7.7.2',
+    url = 'https://tauonmusicbox.rocks/',
+    license = 'GPL-3',
+    install_requires=required,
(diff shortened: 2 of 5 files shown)


stkw0:
It should be ready now :)

Taiko2k:
Does this work, like have you tested this? Also don't mind my ignorance but could you explain to me what setup.py does in this instance?

stkw0:
> Does this work, like have you tested this? Also don't mind my ignorance but could you explain to me what setup.py does in this instance?

Yes, I have been testing it. There is still room for improvement. Currently the setup.py allows to easily build and install the t_module in the standard path. This allows tauon.py script to find t_module when importing it. I saw that for Linux it was imported manually, I don't know if the issue is still pressent but seems to work. 

In the package I made for tauon the assets, themes and templates are installed in /usr/share/TauonMusicBox. The python part is installed in the standard python paths and the tauon.py is installed in /usr/bin/tauon.py.

Some things that can be improved but I am not sure about is to add input.txt and assets in setup.py so the installation process is shared across different Linux distributions.

Taiko2k:
FYI the input.txt is a template and not intended to be modified, only the copy that gets put in the user data folder. So I guess it goes with other static assets.

Also consider the compile translations script, dunno if that should be automatic.

Also note there may be some dependencies missing from the requirements txt, such as pygobject. Im not sure how system provided python packages vs pip provided packages should work.

stkw0:
Thanks, in the draft for the package I forgot about the translations, tomorrow I will add it as long as other misc files like the desktop file and such, but that's an issue of package manager side. What I was also considering was adding the possibility to load the phazor.so shared library from `/usr/lib{64}` instead of `install_directory + /lib`, for now I create a symlink to `install_directory + /lib/`. 

About the other dependencies, I think the standard approach (from user point of view) is to let the package manager handled the dependencies to avoid  needing pip or any other solution requiring two (or more) system-wide package managers simultaneously. From the development perspective I think usually what is done is to create a virtual environment isolated from the rest of the system. This way both things can exist at the same time without colliding with one another.

stkw0:
At the end also changed how the phazor.so is loaded:` aud = ctypes.cdll.LoadLibrary("libphazor.so")`
By removing the hardcoded path we ensure that it search the library on the standard LD_LIBRARY_PATH avoiding workarounds while maintaining compatibility. First it checks from `install_directory/lib/libphazor.so` anyway to ensure compatibility, but maybe it could be removed. I don't know how the phazor components is handled on Windows.

Taiko2k:
Its exactly the same path on Windows, just the `lib/libphazor.so` file is actually a compiled dll.

stkw0:
I see. So the .so file can actually be a .dll, right? In any case with the current implementation it is compatible with the previous behavior, but if it fails to find the .so  then it looks at the standard paths.

stkw0:
The Gentoo package has been using this patch for a while. I had no issues with it nor received any bugs about it. Could it be merged?

PR #1137, opened 2024-04-26
Tweaked Discord album art implementation to make it more reliable. Closes #1136.

### LINKED ISSUE #1136: Get album art from `musicbrainz_albumid` tag
The album art lookup seems to fail often even on albums which have existing album art in Musicbrainz. Most of my songs are tagged with Picard, so looking up the MBID from the `musicbrainz_albumid` tag would be a better way of searching for album art.

More information here: https://picard-docs.musicbrainz.org/en/variables/tags_basic.html

G2-Games:
It now uses both `release_group_id` and `release_id`, which makes it significantly more reliable.

### DIFF
--- a/t_modules/t_main.py
+++ b/t_modules/t_main.py
@@ -23820,7 +23820,7 @@ def get_album_art_url(tr):
     if not artist:
         return
 
-    album_id = None
+    release_id = None
     release_group_id = None
     if (artist, tr.album) in pctl.album_mbid_cache:
         mbid = pctl.album_mbid_cache[(artist, tr.album)]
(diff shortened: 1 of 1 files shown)


G2-Games:
Closes issue #1136

t_modules/t_main.py
```diff
@@ -23820,53 +23820,91 @@ def get_album_art_url(tr):
...
+    if image_data:
+        for image in image_data["images"]:
             if image.get("front") and "250" in image["thumbnails"]:
                 pctl.album_mbid_cache[(artist, tr.album)] = mbid
                 url = image["thumbnails"]["250"]
                 if url:
                     print("got mb image url for discord")
                     pctl.mbid_image_url_cache[mbid] = url
                     return url
+    else:
```
Taiko2k, line 23906: This else path isn't cached?

t_modules/t_main.py
```diff
@@ -23820,53 +23820,91 @@ def get_album_art_url(tr):
...
+            image_data = response.json()
+            mbid = release_id
+        except (requests.RequestException, ValueError):
+            #print("no image found for album id")
+            pctl.album_mbid_cache[(artist, tr.album)] = None
+
+    if image_data:
+        for image in image_data["images"]:
             if image.get("front") and "250" in image["thumbnails"]:
                 pctl.album_mbid_cache[(artist, tr.album)] = mbid
```
Taiko2k, line 23900: Its no longer clear what this `mbid` named variable is or does, we should refactor this specific variable away.

Taiko2k (changes requested):
I'm not sure the caching mechanism still works as intended with this revision.

The goal is that we want to minimize any unnecessary repeated calls. Caching both results and failures. To this end I created the dicts `pctl.album_mbid_cache` and `pctl.mbid_image_url_cache`. These can be changed as needed, though your change has implications for how the former of the two works. 

Could you revise the PR to ensure we don't make an API call with the same data twice. (within in the same session). You can revise the caching mechanism and associated variables as needed.

G2-Games:
Good points, I will revise my implementation

G2-Games:
That should fix the caching issues I think

PR #1155, opened 2024-05-21
Add High Contrast Dark theme

CryHavoc21:
Added High Contrast Dark theme. Primarily based on Windows' High Contrast Dark theme 

Example: 
image

Unfortunately, it looks like this disabled button background color and this timestamp for non-playing songs is not affected by the theme. I suspect these colors might be hardcoded

image


Taiko2k:
I think those elements *shouldn't* be hardcoded. There might be a bug but I believe the keys for those in the theme file are the following:

 - Tracklist un-highlighted time display is `track time`
 - Text of an unseleted playlist tab is `tab text`

Try double check what you have those set to.

CryHavoc21:
I see, I found the problem, the named keys for some of the keys had been accidentally changed

I've put them back to normal and fixed the indentation of them. Here's a preview of the new theme: 
image

PR #1164, opened 2024-06-02
replace isounidecode with unidecode

stkw0:
isounidecode don't have any new commits (nor releases) since 7 years ago. Also, seems that distros don't tend to package it (at least Gentoo, Debian and Fedora). 
Unidecode seems to be API compatible and work out of the box and is more active.

### DIFF
--- a/extra/win-requirements.txt
+++ b/extra/win-requirements.txt
@@ -8,7 +8,7 @@ requests
 beautifulsoup4
 PlexAPI
 musicbrainzngs
-isounidecode
+unidecode
 infi.systray
 mutagen
 natsort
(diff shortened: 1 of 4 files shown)


Taiko2k:
I think this is good, but its not exactly the same API. `unidecode()` from isounidecode returns `bytes` where as unidecode returns `str`. So you just need to remove the `.decode()` from of any usages of `unidecode().decode()` in those two source files.

As a basic test just perform a global search in Tauon, if you can get any results to show then there should be no major problems.

Taiko2k:
Also perhaps change `"GPL-2.0"` to `"GPL-2.0+"`, to be clear we are using the "or later" clause.

stkw0:
Done, the search seems to work :)

PR #1177, opened 2024-07-06
Search through lyrics

RSKYS:
(Work in progress)


RSKYS:
There are currently many changes; I'm not adding this until next release.
I rebase my patches against stable releases, and it is 7.8.0; which there were too many changes like unicode and whatever else heck in t_main.py.
I decided to take my time and not to rush for this.

PR #1178, opened 2024-07-07
Improvement to Discord RPC

RSKYS:


Taiko2k:
The rate limit discord imposes is 15 seconds, so lowering the loop tick rate will either increase chance of getting rate limited or have no effect. 

If we really want to tighten the timings, perhaps after 15 seconds since last update, then spinlock on 2 seconds (perhaps also with a `pctl.a_time > 5` condition to break (`pctl.a_time` is how long the current track has been playing, I know, not a very descriptive variable name))

RSKYS:
The main problem I had was if I press "/" for quiet long and change too many tracks, it will get stuck on some specific track, but seem this fixed it
I can also revert 15 back, but I had no issue changing too many tracks with it.

Cheers

Taiko2k:
Are you sure it updates faster than 15? Most documentation sources say it shouldn't.

Taiko2k:
Actually looking at newer discord docs the new rate limit might actually be 4s, so, maybe we could put try this as 5s to be safe.

Also could you change the line `if pctl.playing_time > 4 or current_state != 1:` to `if pctl.a_time > 4 or current_state != 1:` for me.

PR #1179, opened 2024-07-09
enhance import

octoamit:
This pull request enhances the Jellyfin Collection import by updating the GET
request, increasing the import speed, and further populating TrackClass
attributes.

The goal is to decrease the time and system resources for importing /
regenerating a Jellyfin Collection, and more fully populate track information.

The Jellyfin source code indicates the "/Users/{userId}/Items" format for
the Items endpoint is obsolete. Currently it is "Kept for backwards 
compatibility".
https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Api/Controllers/ItemsController.cs#L625
To avoid issues in the future if or when it is removed, I've updated the 
request to be consistent with the /Items endpoint format used in the Jellyfin 
API documentation.
https://api.jellyfin.org/#tag/Items

The "filter": "music" request parameter currently has no effect. The entire 
media library is returned in the response. So I've replaced it with the
"mediaTypes": ["Audio"] parameter which provides the desired result of only
returning music and playlists. Playlists are always of mediaType Audio, even 
when they include only video or a mix of audio and video. Depending on the media
library size and consistency, this change can drastically decrease the time 
required for import and/or regeneration.

The "fields" parameter for the Items endpoint allows you to "Specify additional 
fields of information to return in the output." The TrackClass attributes 
populated by the get_info() method can be more directly, fully, and quickly 
populated within the Jellyfin import loop by including "MediaSources" in the
"fields" parameter list and using it in the Jellyfin import loop.

While testing, I've found it is much much faster, and less cpu and network 
intensive, to fully populate the TrackClass attributes when initially iterating
through the sorted and grouped json response, instead of calling the get_info() 
method to reiterate over the tracks again, each time making a request to the 
server.

Therefore, I added "MediaSources" to the "fields" parameter list, moved the
items populated in the get_info() method to the Jellyfin import loop, and
removed the get_info() method and calls to it. Additionally, I have removed the
lines in the Jellyfin import loop related to deducing the parent folder name
and path, because they are already provided by the "MediaSources" field and
collected in the lines moved from the get_info() method to the Jellyfin import
loop.

I also added lines to fully populate every TrackClass attribute I could. I added
the "DateCreated" and "People" fields parameter in order to populate the 
modified_time and composer TrackClass attributes, respectively. I was also able
to populate the "size" and "comment" attributes from the MediaSources field.

Including the query parameter "enableImages": 'false' will reduce the size of 
the response payload to various degrees depending on the media library. 
This does not impact cover art displaying because the GET response
will still include the "AlbumPrimaryImageTag" used by the get_cover method.

I ordered the attributes to correspond to / appear in the same order as they 
appear in the TrackClass class.

I have tested the changes to ensure it functions correctly and I have not had
any issues with these changes. All the metadata is displayed correctly and the
time and system resources required for import and/or regeneration has 
drastically decreased.

I could not find a way to reduce the size of the response payload by limiting 
the response to return only the specific attributes required to populate 
the TrackClass class. Currently it appears there is no such option, so most of 
the response payload ends up discarded. There could be some potential 
improvement by processing the response as it is received, but even if it is 
possible it would require many more changes.

### DIFF
--- a/t_modules/t_jellyfin.py
+++ b/t_modules/t_jellyfin.py
@@ -351,48 +351,6 @@ class Jellyfin():
             self.pctl.multi_playlist.append(self.tauon.pl_gen(title=p['Name'], playlist=playlist))
             self.pctl.gen_codes[self.tauon.pl_to_id(len(self.pctl.multi_playlist) - 1)] = f"jelly\"{p['Id']}\""
 
-    def get_info(self, fast=False):
-        dones = []
-        i = 0
-        for p in self.pctl.multi_playlist:
-            for t in p[2]:
-                tr = self.pctl.g(t)
-                i += 1
-                if i % 1000 == 0:
-                    print(i)
-                if tr.file_ext == "JELY":
-                    if tr.fullpath and fast:
-                        continue
-                    if tr.url_key not in dones:
-                        dones.append(tr.url_key)
-
-                    # Get media info
-                    response = requests.get(
-                        f"{self.prefs.jelly_server_url}/Items/{tr.url_key}/PlaybackInfo",
-                        headers={
-                            "Token": self.accessToken,
-                            "X-Application": "Tauon/1.0",
-                            "x-emby-authorization": self._get_jellyfin_auth()
-                        },
-                        params={
-                            "UserId": self.userId,
-                        }
-                    )
-                    if response.status_code == 200:
-                        try:
-                            d = response.json()
-                            tr.fullpath = d["MediaSources"][0]["Path"]
-                            tr.filename = os.path.basename(tr.fullpath)
-                            tr.parent_folder_path = os.path.dirname(tr.fullpath)
-                            tr.parent_folder_name = os.path.basename(tr.parent_folder_path)
-                            tr.misc["container"] = d["MediaSources"][0]["Container"].upper()
-                            tr.misc["codec"] = d["MediaSources"][0]["MediaStreams"][0]["Codec"]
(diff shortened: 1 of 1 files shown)


Taiko2k:
Looks good. 

Just found one issue, with:
`nt.misc["container"] = track.get("Container").upper()`
I just get a None type error. But this works for me:
`nt.misc["container"] = track.get("MediaSources")[0]["Container"].upper()`

Ive gone ahead and committed this adjustment to the master branch. I wonder why the former worked for you and not me. Can you test this change?

octoamit:
Sure, I tested this change; it works great and yields what is expected for the container.

Sorry for the trouble. I had assumed the former "Container" (appearing 20th in the Items response) was the same as the "Container" in the MediaSources dictionary. I had changed it to use the former because at first I was trying to avoid including MediaSources parameter in order to reduce the size of the response, but the MediaSources parameter is necessary to fully populate the TrackClass attributes.

It appears that the former "Container" comes from the files' metadata whereas the latter comes from the filenames. When using the former "Container", I did not get a None type error, but I did notice some of my audio files were showing 'mov,mp4,m4a,3gp,3g2,mj2' for "Container" instead of m4a. This string 'mov,mp4,m4a,3gp,3g2,mj2' matches the "format_name" field in those files' metadata. 

```
$ ffprobe -v error -hide_banner -show_format example.m4a | grep format_name
format_name=mov,mp4,m4a,3gp,3g2,mj2
```

PR #1182, opened 2024-07-11
Specify Audio MediaStreams

octoamit:
In the Jellyfin json response there can be more than one "MediaStreams" dictionary. In my audio files the first (or only) one is usually audio, but I've discovered in some cases the first stream is cover art.

This pull request ensures that the TrackClass attributes populated from the MediaStreams dictionary is from the audio media stream rather than some other. I was also sure to include a default in the event one of the attributes were missing.

I've also added a fallback empty string for the nt.date and nt.container attributes in the event they are missing, as well as changing the enableImages query parameter from 'false' to False (it works either way, but the latter seems more "correct"). All the other attributes already have a fallback, or do not need one because it would otherwise not exist (e.g. file path, size, modified date).

By the way, setting enableImages to False will not impact cover art displaying because the GET response will still include the "AlbumPrimaryImageTag" used by the get_cover method. By setting enableImages to False I am able to reduce the size of the json response by ~30MB for my library, which isn't much, but does further reduce the size of the json response.

I have tested these changes to ensure that it functions correctly. 

These changes will pull the data from the first audio stream. I suppose technically there could be more than one audio media stream in a music file. At this time, I'm not sure what should be done in the case of a second audio stream in a music file but, while I've seen multiple audio streams in video files, it seems like it would be extremely unusual and rare for music files.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,9 +5,9 @@ Wanna contribute? Theres a few ways you can:
 
 ## Questions
 
-You feel free to ask questions as a new Github issue. Please do not comment in an existing or old issue unless you're the author or you're sure its directly releated.
+You feel free to ask questions as a new Github issue. Please do not comment in an existing or old issue unless you're the author or you're sure its directly related.
 
-You can also use the github discussions feature. 
+You can also use the github discussions feature.
 
 If you use Discord it may be faster to ask on the official server.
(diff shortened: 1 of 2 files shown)


octoamit:
I noticed and fixed a few misspellings in the CONTRIBUTING.md file too.

PR #1187, opened 2024-07-14
and is &

RSKYS:


Taiko2k:
While this may fix some use cases I anticipate it may break others. Like what if what a user is searching actually includes the word "and". A more correct approach would be to perform a search comparison of both forms somehow but I would have performance reservations about that.

RSKYS:
My case is limited to these 5 variables in search anyway, but I might understand your point of view, it overwrites default search result of tracks with & too.
I'll think about it sometime later.

t_modules/t_main.py
```diff
@@ -25662,6 +25664,11 @@ def worker2():
                                 search_string_cache[
                                     track] = title + artist + album_artist + composer + date + album + genre + sartist + filename + stem
 
+                        if "and" in s_text:
```
Taiko2k, line 25667: This should go further up outside of the search loop

t_modules/t_main.py
```diff
@@ -25662,6 +25664,11 @@ def worker2():
...
+                            search_string_cache.clear()
```
Taiko2k, line 25668: These need to be called when switching *and_mode*, so you will need another variable to keep track of the previous state and only clear the cache when the state changes. Again this can go outside of the search loop.

t_modules/t_main.py
```diff
@@ -25673,8 +25680,18 @@ def worker2():
                                         s_text = s_cn
 
                         if dia_mode:
-                            title = unidecode(title)
+                            if and_mode and "&" in title:
```
Taiko2k, line 25683: These can be placed in a single `if and_mode` block

Taiko2k (changes requested):

t_modules/t_main.py
RSKYS: Yea yea yea, I pushed these fast, was still thinking, you told me already, gonna come back.

RSKYS:
I'm not interested to write & thingy, it requires minor changes compared to what I did, but there's been few days I'm thinking if I should really contribute to the project or not anymore.
Do it yourself if you wanted, I stay out of  python.
Maybe maybe if I saw something worthy; right now I might sometimes contribute to phazor, there are some mpt things I'd put to practice later.

PR #1201, opened 2024-07-30
Few fixes for phazor

RSKYS:


src/phazor/phazor.c
```diff
@@ -1370,9 +1370,8 @@ static int pipe_update(struct spa_loop *loop, bool async, uint32_t seq,
                      const void *_data, size_t size, void *user_data){
 
         pw_stream_disconnect(global_stream);
-        pipe_connect(loop, async, seq, _data, size, user_data);
+        return pipe_connect(loop, async, seq, _data, size, user_data);
         printf("UPDATE SAMPLERATE PIPE TO %d\n", pipe_set_samplerate);
```
Taiko2k: This line is unreachable now?

src/phazor/phazor.c
```diff
@@ -1421,7 +1420,6 @@ void connect_pulse() {
 
     //printf("ph: Connect device\n");
 
-    ma_context_config c_config = ma_context_config_init();
```
Taiko2k, line 1424: Can you explain this one to me? Is it not needed?

RSKYS: Seems it already had initiate value and therefore, it works fine without it

src/phazor/phazor.c
RSKYS: Since it's int function, it should return value there, that's why I returned last function to ignore possible structure warn.

Taiko2k: But the printf statement is pointless there now

src/phazor/phazor.c
Taiko2k: Maybe we would leave it there if theres no error

RSKYS: Okay, I'll push it again, it leaves a warn but if you wanna be sure then oki doki :)

Taiko2k: Whats the warning?

src/phazor/phazor.c
RSKYS: Do you want me to write printf above it?

Taiko2k: Just remove it then

RSKYS: I simply return 0 maybe

src/phazor/phazor.c
RSKYS:
```
src/phazor/phazor.c:1424:23: warning: variable ‘c_config’ set but not used [-Wunused-but-set-variable]
 1424 |     ma_context_config c_config = ma_context_config_init();
      |                       ^~~~~~~~
```

PR #1247, opened 2024-10-12
Made most chars localizable 

Lorsoen:
As in the title, plus some more extras along the way (mentionned below),

I've been satisfied with Tauon's rich functionnality but I think it could benefit from a better localization work (notably the GUI who still surprinsingly have a lot of _english remaining in GUI_). 

This PR intends to pave the way for that + A _french translation overhaul_ done in parallel of this PR, aiming to fix layout overlap issues, walls of text and inconsistencies. The only non-touched chars are the console prints.

The extras:

- _Most links are now unlocalizable_ to avoid malicious intents (beacuse Weblate). https:// was added in localizable chars that have links to add better context for Weblate translators.
- ~~The _GPL line was modified_ beacause I couldn't make the 2 chars between the GPL link translatable (I can leave the "About" section untranslated if you wish to).~~


t_modules/t_main.py
```diff
@@ -13971,7 +13971,7 @@ def load_xspf(path):
             console.print("-- -- Album: " + track['album'], level=2)
 
     if missing > 0:
-        show_message('Failed to locate ' + str(missing) + ' out of ' + str(len(a)) + ' tracks.')
+        show_message(_('Failed to locate ') + str(missing) + _(' out of ') + str(len(a)) + _(' tracks.'))
```
Taiko2k: Something like this would better be rewritten as a format string, you can't just translate parts of a sentence

t_modules/t_jellyfin.py
```diff
@@ -397,16 +397,16 @@ def ingest_library(self, return_list=False):
...
             self.scanning = False
             return
 
         if response.status_code == 200:
             print("Connection successful, storing items...")
 
             # filter audio items only
-            audio_items = list(filter(lambda item: item["Type"] == "Audio", response.json()["Items"]))
-            playlist_items = list(filter(lambda item: item["Type"] == "Playlist", response.json()["Items"]))
+            audio_items = list(filter(lambda item: item["Type"] == "Audio"), response.json()["Items"])
```
Taiko2k: Logical change of non user facing strings

t_modules/t_jellyfin.py
```diff
@@ -517,7 +517,7 @@ def set_favs(d):
                     if star is None:
                         pass
                     else:
-                        star = [star[0], star[1].replace("L", ""), star[2]]
+                        star = [star[0], star[1].replace("L"), "", star[2]]
```
Taiko2k: Logical change

t_modules/t_main.py
```diff
@@ -17788,17 +17788,17 @@ def export_stats(pl):
     #     line = line.rstrip("; ")
     #     line += "\n\n"
 
-    line += "\n-------------- Top Artists --------------------\n\n"
+    line += _("\n-------------- Top Artists --------------------\n\n")
```
Taiko2k: Probably should be a format string

t_modules/t_main.py
```diff
@@ -30281,7 +30281,7 @@ def topchart(self, x0, y0, w0, h0):
         y = y0 + 100 * gui.scale
 
         # . Limited width. Max 13 chars.
-        if self.button(x, y, _("Randomise BG")):
```
Taiko2k: Tauon is intended to use British spelling by default, this probably needs to be split off to en-US translation

t_modules/t_main.py
```diff
@@ -17690,19 +17690,19 @@ def export_stats(pl):
     total_size = sum(seen_files.values())
 
     stats_gen.update(pl)
-    line = 'Playlist:\n' + pctl.multi_playlist[pl][0] + "\n\n"
-    line += 'Generated:\n' + time.strftime("%c") + "\n\n"
-    line += 'Tracks in playlist:\n' + str(tracks_in_playlist)
+    line = _('Playlist:\n') + pctl.multi_playlist[pl][0] + "\n\n"
```
Taiko2k: Format controls should probably be left out of the translate string

t_modules/t_jellyfin.py
```diff
@@ -527,7 +527,7 @@ def set_favs(d):
             self.scanning = False
             return playlist
 
-        self.pctl.multi_playlist.append(self.tauon.pl_gen(title="Jellyfin Collection", playlist=playlist))
+        self.pctl.multi_playlist.append(self.tauon.pl_gen(title="Jellyfin Collection"), playlist=playlist)
```
Taiko2k: Logical change?

t_modules/t_main.py
```diff
@@ -8460,16 +8460,16 @@ def go():
                     if mb > 90:
                         break
                     if mb % 5 == 0:
-                        show_message(f"Downloading... {mb}/80MB")
+                        show_message(_(f"Downloading... {mb}/80MB"))
 
             except Exception as e:
-                show_message("Download failed", str(e), mode="error")
+                show_message(_("Download failed", str(e)), mode="error")
```
Taiko2k, line 8466: Translate of a dynamic string?

t_modules/t_main.py
```diff
@@ -40924,13 +40924,13 @@ def render(self):
             gui.update_on_drag = True
 
         if insert is not None:
-            radios.insert(insert, "New")
+            radios.insert(insert, _("New"))
```
Taiko2k: This is a non user facing string actually

t_modules/t_main.py
```diff
@@ -9506,14 +9506,14 @@ def sub_get_album_thread():
 
 def koel_get_album_thread():
     # if prefs.backend != 1:
-    #     show_message("This feature is currently only available with the BASS backend")
+    #     show_message(_("This feature is currently only available with the BASS backend")
```
Taiko2k, line 9509: While commented out strings probably don't matter, this code is now invalid

Taiko2k (commented):
- This PR is quite large, it would be preferred to make multiple smaller PRs to make code review easier.
- It looks like you might have used a script to make sweeping changes, this is generally a no-no especially if you don't manually check the output.
- Cant really change the po/pot files in a pull *request* as it will make weblate out of sync and wont merge when it comes time to. Just leave those out and it will update it in time.
- Files outside of t_main.py aren't currently programmatically set up for translation (it is what it is), at the least I think other py files would need a dummy `_()` function, does this PR run without errors?

Lorsoen:
First of all, thank you a lot for your detailed review :+1: 

I've pushed changes to adress the things mentionned (exept one where i'm clueless, see conversations).

t_modules/t_main.py
Lorsoen: I don't get it ? There's no value to call here.

Lorsoen:
> Files outside of t_main.py aren't currently programmatically set up for translation (it is what it is), at the least I think other py files would need a dummy _() function, does this PR run without errors?

I've tested extensively on my translation branch, with a modified update-translation.py script who includes all modules with GUI strings that need translations. It worked without crashes so far.

Lorsoen:
Working example of a translated char being called from an eternal module (on my french_translation branch, very similar to this one) :

https://github.com/user-attachments/assets/80eeb30c-92d5-44d0-892d-9fbe2745d14c

Lorsoen:
Just noticed this in t_main : 

Copie d'écran_20241013_164748

This won't work with a lot of languages. Should the plural logic be stripped and to something like : "Track(s)" or "Track.s" ?

Taiko2k:
> Just noticed this in t_main : 
> 
> 
> 
> Copie d'écran_20241013_164748
> 
> 
> 
> This won't work with a lot of languages. Should the plural logic be stripped and the char being something like : "Track(s)" or "Track.s" ?

Not sure what the best practice is but I'd suggest changing the logic to two different strings to translate the full words "Track" and "Tracks", rather than trying to append the "s"

Lorsoen:
I've made a change on localizable strings and the way it handles variable calls.

Most of the _prefixes_ (f-string) and _% operators_ were _replaced by ".format"_ instead. 
(Replacing prefixes was mandatory, since those don't work with python's "get text" function, making strings using them unlocalizable).

Even tho it is admittedly less readable on coder side, it comes with two advantages for translators:
- **Better contextualisation for translators on Weblate.** 
Sometimes chars like those may be hard to contextualize without picking the code or if you're Python illiterate:
`"%d listening along"` (Is it a name or a number ?)
This method allows variable calls to have a bit more explicit names for translators:
`"{N} listening along"`

- **Better translation flexibility.**
% operators were limited by the fact that, in a single string, the order where variables are called is always the same (afaik).
As an example, a complex string like this:
`"Start of location {N} of {T} ({Nb} albums)"`
Could easily shift the values if needed or desired:
`"({Nb} albums) Start of location {N} of {T}"`
This allows translators to go a bit wilder and writing phrases that feels natural in their language, without being limited by value call order.

Lorsoen:
One last area that lacks translation is the _default playlist_. 

I have no idea how to make it translatable, other than _moving the language file loading_ bit before any localizable string. This involves moving around some loading components to avoid unset variables being called too early.

Other thant that, it's _ready for review_, again.

t_modules/t_main.py
```diff
@@ -23069,13 +23070,13 @@ def export_database():
...
-    show_message("Export complete.", "Saved as: " + path, mode='done')
+    show_message(_("Export complete."), _("Saved as: ") + path, mode='done')
 
 
 def q_to_playlist():
-    pctl.multi_playlist.append(pl_gen(title="Play History",
+    pctl.multi_playlist.append(pl_gen(title=_("Play History"),
                                       playing=0,
-                                      playlist=list(reversed(copy.deepcopy(pctl.track_queue))),
+                                      playlist=lisft(reversed(copy.deepcopy(pctl.track_queue))),
```
Taiko2k, line 23079: Is this a typo?

t_modules/t_main.py
Taiko2k: I mean maybe instead of `+ _("-------------- Top Artists --------------------")` we should do `+ f"-------------- {_("Top Artists")} --------------------"` for this section to make less cumbersome to translate.

t_modules/t_main.py
```diff
@@ -40928,9 +40944,9 @@ def render(self):
             if self.drag in radios:
                 radios.remove(self.drag)
             else:
-                toast(_("Added station to: " + pctl.radio_playlists[pctl.radio_playlist_viewing]["name"]))
+                toast(_("Added station to: ") + pctl.radio_playlists[pctl.radio_playlist_viewing]["name"])
 
-            radios[radios.index("New")] = self.drag
+            radios[radios.index_("New")] = self.drag
```
Taiko2k: Logical error. This is an internal / non user facing string btw

t_modules/t_main.py
```diff
@@ -8998,7 +8998,7 @@ def getsongs(index, folder_id, name, inner=False, parent=None):
     #     try:
     #         a = self.r("getIndexes")
     #     except:
-    #         show_message("Error connecting to Airsonic server", mode="error")
+    #         show_message(_("Error connecting to Airsonic server", mode="error")
```
Taiko2k: No closing bracket

t_modules/t_main.py
```diff
@@ -9025,7 +9025,7 @@ def getsongs(index, folder_id, name, inner=False, parent=None):
     #
     #         except json.decoder.JSONDecodeError:
     #             print("Error reading Airsonic directory")
-    #             show_message("Error reading Airsonic directory!", mode="warning")
+    #             show_message(_("Error reading Airsonic directory!", mode="warning")
```
Taiko2k: No closing bracket

Taiko2k:
Apologies for the delay in merging this, but are you able to rebase this?

Lorsoen:
No problem :+1:  Should be good to go.

C0rn3j:
@Lorsoen Just a small thing - 

You might want to set your editor to display whitespace and/or obliterate trailing whitespace on saves.

This PR introduced a couple changes like these (this is my commit, hence it's being deleted) - 

image

Unfortunately, GitHub has a bug and does not show it in the diff for this PR in the web UI - 

image

C0rn3j:
Ah, a bigger issue though is that you reverted some of the other changes during your rebase, whitespace notwithstanding for example the docstring here.

I'll fix it up.

image

PR #1253, opened 2024-10-19
Add basic linter rules and partially type Phazor, Spotify and Webserve

C0rn3j:
Follow up to #1251.

There's a couple things to go through:

* Single vs Double quotes - it would be nice to consistently use one formatting
The codebase uses quotes pretty much randomly:
image  
I've added a linter rule that ensures single quotes everywhere (docstrings remain double quoted, as per PEP recommendations).  
If double quotes are desired instead, let me know and I'll swap it.

* Spaces vs tabs - It would be nice to use tab indents, as that will allow everyone to view the code at the indentation width they prefer  
Using tabs goes directly against PEP8. In my opinion, this PEP8 recommendation is nonsense today, as Python handles indentation amazingly for both, and tabs allow people their own indent preference.  
Everyone uses something else - I prefer 2-width, htop uses 3-width, TMB uses 4-width and Linux uses 8-width.  
Tabs allow people to set their editors to display what they want.  
Editors had poor support for setting indent width in the past, which is why we're historically stuck with things like the PEP8 recc.  
Currently, the PR has spaces in the linter, as that's what the codebase uses.  
If conversion would be agreeable, it's not necessary to convert the entire code base at once (which would be impossible as indents are not consistent even now, this PR fixes one such case, and using tabs has considerations for their dynamic sizing), but Python allows separate files to have different indents from other files, or even within the same file for different functions, one just can't mix them for code within the same function block.

* Removed wildcard imports  
I was able to remove the wildcard imports from phazor, but was not able to test properly due to build system issues below, so it might be a good idea to test that it still runs on both Linux + Windows.

* Platform checks  
I have swapped `tauon.msys` checks, which I understand are currently used as a kinda check for both win32 + windows terminal that does not support Unicode well (is this still the case?), for `sys.platform == "win32"` checks in the Phazor files.  
I think that at least in here that will work correctly, but I it would be nice to take a look that I did not mess it up.  
This was necessary to make type checkers to work, as they otherwise try to include the Windows code when on Linux.

* Current build system  
This is a ticket for another PR, but Tauon currently relies on the old `setup.py` over `pyproject.toml` (which this PR uses to add linter rules).  
I believe Phazor needs to somehow be built first before launching tauon.py, but running `compile-phazor.sh` just errors on missing miniaudio... is there a proper build script somewhere in the project files? I am aware I can just copy build steps from AUR PKGBUILDs.  
Would be nice if this built outright from `pyproject.toml`.

The rest is just adding more typing, and adding some missing variables to the Tauon class.  
Would have added more things, but it feels like enough changes for one PR and it'll probably be best to figure out how to properly build and test before I move on :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,10 +5,9 @@ target-version = "py39"
 line-length = 120
 
 [tool.ruff.lint.flake8-quotes]
-inline-quotes = 'single'
+inline-quotes = 'double'
 docstring-quotes = 'double'
 
 [tool.ruff.format]
-quote-style = 'single'
+quote-style = 'double'
 indent-style = 'space' # Change to tabs?
-
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,14 @@
+[tool.ruff]
+# Target non-EOL releases at minimum - https://devguide.python.org/versions/
+target-version = "py39"
+# Soft 80 and hard break at 120
+line-length = 120
+
+[tool.ruff.lint.flake8-quotes]
+inline-quotes = 'single'
+docstring-quotes = 'double'
+
+[tool.ruff.format]
+quote-style = 'single'
+indent-style = 'space' # Change to tabs?
+
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,4 +10,4 @@ docstring-quotes = 'double'
 
 [tool.ruff.format]
 quote-style = 'double'
-indent-style = 'space' # Change to tabs?
+indent-style = 'tab'
--- a/t_modules/t_main.py
+++ b/t_modules/t_main.py
@@ -1,7 +1,12 @@
-#! /usr/bin/env python3
-# -*- coding: utf-8 -*-
+"""Tauon Music Box
 
-# Tauon Music Box
+Preamble
+
+Welcome to the Tauon Music Box source code. I started this project when I was first
+learning python, as a result this code can be quite messy. No doubt I have
+written some things terribly wrong or inefficiently in places.
+I would highly recommend not using this project as an example on how to code cleanly or correctly.
+"""
 
 # Copyright © 2015-2024, Taiko2k captain(dot)gxj(at)gmail.com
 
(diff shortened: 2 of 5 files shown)


Taiko2k:
Tabs, single quotes, build systems... why this has to be the WORST pull request I've ever seen 😁 .

OK OK,

 - Quotes: I prefer double quotes. Its what I'm used to. I used to use the black formatter for a while to I tend to lean toward its formatting style.
 - Tabs: You might be able to convince me on this one, but I'm not sure on committing a rewrite of every file, I feel its bad practice since it might mess up git etc.
 - No wildcard: t_main.py uses a lot of imports from t_extra so I don't see that as being feasible there.
 - Platform check: That one there should be OK, but in general, since MSYS2 kind of emulates Linux, some things behave like Linux when running under that, and Tauon has some major different code paths for that, hence the distinction. Although just to note, code paths for full native win32 mode haven't been maintained in a long time and are likely broken.
 - Build system: Someone submitted a PR for that setup.py, but I don't use it personally. Feel free to change it. Yeah basically the most official way to build is what the PKGBUILD does.

C0rn3j:
> Quotes: I prefer double quotes. Its what I'm used to.

Swapped.  Will apply it for the files I touch in future commits.

I prefer single personally, but that's only because I get flashbacks from shell interpreting things within double quotes. Got no bone to pick with Python there.

> No wildcard: t_main.py uses a lot of imports from t_extra so I don't see that as being feasible there.

Oh that is eventually hopefully feasible too, but I was only talking about the specific phazor file being rid of them, as I am not sure I 100% correctly caught everything when fixing it. 

I tried my best, but the current code throws 90 type errors\* in the phazor file, so I could have overlooked it despite combing it over repeatedly.  

They are errors discovered by adding the types, I am not fixing them, for now anyway, just adding the type checking.

They're mostly assumptions that the used variables are not None, despite the fact they could be as the called code paths/functions do not guarantee them not to be.

> Platform check: That one there should be OK, but in general, since MSYS2 kind of emulates Linux, some things behave like Linux when running under that, and Tauon has some major different code paths for that, hence the distinction.

Hm, I suppose adding the win32 check AFTER msys check would satisfy the type checker in cases where it's not feasible to swap it directly like the PR does, and wouldn't mess the logic otherwise.

Will have to think on that if I keep chewing through this, maybe there's a better way where the variable is kept and the checker understands it's not possible to hit on Linux...

> Build system: Someone submitted a PR for that setup.py, but I don't use it personally. 

I might PR an improvement for that in the future then, as it would be nice if the Python build system could build the project instead of having to do it manually.

And it would make CI tests more feasible...

If you want to take a peek, you can checkout how sc-controller does things. Note it does use [setup.py](https://github.com/C0rn3j/sc-controller/blob/python3/setup.py) for C stuff still, as it does not seem like there is a way to do it via  [pyproject.toml](https://github.com/C0rn3j/sc-controller/blob/python3/pyproject.toml) for now.

> Tabs: You might be able to convince me on this one, but I'm not sure on committing a rewrite of every file, I feel its bad practice since it might mess up git etc.

GitHub at minimum is smart enough to not put whitespace-only changes into blame, so it should be fine.  
See https://github.com/C0rn3j/sc-controller/blame/732f6c5b2fba1ded249a96b152ed464efc97090a/scc/drivers/ds5drv.py

PR #1277, opened 2024-11-15
t_main: Rework imports - put on top, remove some star imports and dupes

C0rn3j:
Also adds explicit Exceptions everywhere.

Merge is dependent on the 3 linked PRs, otherwise it will crash as the SDL star imports from elsewhere will mess it up.

I am not 100% sure I caught all the SDL imports, as my tooling breaks trying to parse the 50Kloc file, but I tried my best to.  
It at least crashes with a very clear `undefined` trace when it does so.

### DIFF
--- a/t_modules/t_main.py
+++ b/t_modules/t_main.py
@@ -328,7 +328,7 @@ def whicher(target):
             return "bin/" + target in r
         else:
             return shutil.which(target)
-    except:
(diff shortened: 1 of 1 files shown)


C0rn3j:
No clue what conflict I just created with a force push that fixed a typo...

PR #1302, opened 2024-11-18
Main module improvements, build improvements and other bits

### LINKED ISSUE #1240: Found XDG-Music (uh no you didn't)
Sorry for the cheeky title but I was looking for a new player since I've never really been happy with the others I've used and I'd never seen this before. So I installed, added my music directory...and waited...and waited...and waited. It's been well over an hour and looking at the track count something is extremely wrong. So I killed it and started digging into what the hell is going on.

Running from a terminal it seems to be ignoring my Music directory and spits out XDG-Downloads and XDG-Music which are both wrong and are pointing to my whole home folder...which has over 20 years of DAW/Studio files on it.

So it's not only ignoring the user selected / defined Music folder but it's also got the XDG_DOWNLOAD_DIR and XDG_MUSIC_DIR environment variables wrong.

C0rn3j:
Supersedes #1285

Apologies for the size of this PR, which will increase still, it could in theory be split into smaller parts if really needed.

* Fix spamming `2021` on launch from t_extra as a random bit of example code was hanging there
* Tear out all remaining start imports
* Put optdepends into requirements_optional.txt
* Move Windows depends into requirements_windows.txt
* Attempt to fix `win-build.sh` for new file paths - untested, though
* Fix trying to load Windows version on non-Windows systems
* Remove seemingly useless macOS code for `tap.stop()`, `tap` is not defined anywhere in the codebase
* Fix `transfer-playtime-to` not copying tags
* Fix `run.sh` not deleting egg build folder
* Make `pyproject.toml` packaging package assets, themes, templates, modules. 
  * Parts of the projects were consolidated under the standard directories - `src/tauon`
  * Modules are now imported as `from tauon.t_modules.t_extra` - I suggest dropping the t_ prefix in the future, now that this is packaged together.
* Remove duplicated alpha_blend function
* Split migration code from `t_main`  into `t_db_migrate`
* Removed a piece of code that was always trying to run with an undefined `enc` variable
* Fix broken pylast import
* Upgrade Gtk to 4.0 in t_main, this complements an earlier fix, 3.0 should no longer be accidentally imported and clash with 4.0
* Converted parts of code from `os` file calls to `Path` from `pathlib`
* Converted first 1/10th of t_main from spaces to tabs
* Fixed `reload_albums()` in `t_main` looking unbound `current_date` and `current_title` to the type checker
* BS4 was used in t_main without being imported
* Put the entire codebase under `logging` module, log all exceptions, colorize:   
image
* Fix error loading Xcursor when variables are unset, which seems to be the default on a Wayland session:  
Before: [Screencast_20241120_195815.webm](https://github.com/user-attachments/assets/5c134b31-e4c9-45a9-a508-45b97db67fed)
After: [Screencast_20241120_195455.webm](https://github.com/user-attachments/assets/dce395aa-2368-434e-8e7d-616a5e8b76c9)   
~~Frankly am not sure if I like the fixed behavior?~~ Now only set when the variables are set.
* Fix mishandled `SDL_WINDOWEVENT_DISPLAY_CHANGED` events - gets rid of the Grrr bug
* Fix up loading database/setting files even when they don't exist
* Fix including `opusfile.h` instead of the correct `opus/opusfile.h` in `phazor.c` - seems the same on Windows and Linux

TODO: 
* Fix #1240, without XDG implemented this code will crash due to the improved packaging clashing with hardcoded paths
* Put in translations and anything else that's necessary into `pyproject.toml` too
  * Advice how to resolve that one welcome.
* The coloring of logs might need to be disabled or fixed for Windows as it uses UNIX stuff at the moment - maybe modern Windows terminals support it though?
* I probably won't be addressing it in this PR, but are all those console.print() commands needed now that the logging facility is used? Should we just migrate to logging and get rid of console.print()? Would have to figure out how to log the current messages though to be able to show them in the UI, as it does that. Maybe just log to a file too and open that file on demand?


Taiko2k:
An asset not being present may be warning level but not critical. The app should try to run no matter what state its in unless user data integrity is at risk.

C0rn3j:
If that asset is not present, it crashes below on SDL trying to use it, but with a much less clearer and uncaught error, so I catch it earlier.

Taiko2k:
Oh that's fine then.

C0rn3j:
I've edited in a note about `reload_albums()` as you'll probably know what to do there much quicker than me

C0rn3j:
https://github.com/Taiko2k/Tauon/blob/c80561e0dd56f036d99251040c0b72c1dc23f046/t_modules/t_main.py#L27222-L27307

Taiko2k:
Oh, `current_date` and `current_title` can/should be defined near the top of the function, next to `current_folder = ""` etc. Currently it logically *just werks* but that should make it more readable.

C0rn3j:
> Currently it logically just werks 

I am not sure about that, before the commit with your suggestion that I just sent, the value will imo always be `None` for the comparison.

C0rn3j:
Another thing - when the module is unable to be imported, a message is thrown to inform the user with what's missing due to the failed import:

```python
	logging.warning("Unable to import pypresence, Discord Rich Presence will be disabled.")
```


But it's not obvious to me what's missing with opencc(something with subtitles I guess) and natsort(Faster sort?):

```python
try:
	import opencc
except ModuleNotFoundError:
	logging.warning("Unable to import opencc, TODO(Martin): <what is lost?>.")

try:
	import natsort
except ModuleNotFoundError:
	logging.warning("Unable to import natsort, TODO(Martin): <what is lost?>.")

```

Taiko2k:
If a value is undefined in Python it will crash, not be `None`. Unless those happen to be shadowing a global variable but would then crash on attempt to assign them.

Taiko2k:
- opencc: The ability to perform search using Traditional Chinese and Simplified Chinese interchangeably.
 - natsort: Used for sorting algorithms. The sort order of a playlist may not be the optimal intention if this is missing.

Taiko2k:
> Upgrade Gtk to 4.0 in t_main, this complements an earlier fix, 3.0 should no longer be accidentally imported and clash with 4.0

This issue prevents the app from starting for me so Ive committed this fix ahead of this PR.

C0rn3j:
Regarding last commit, should that still set even when the env vars are empty? See edited OP too for video demos.

C0rn3j:
This is still not ideal, with the missing dealing with save dirs, but it's already massive, might as well get it in and deal with that next in another PR.

PR #1224, opened 2024-09-02
Ricardo simoes patch 2

Ricardo-Simoes:
Portuguese Portugal (pt_PT) 100% translated

PR #1305, opened 2024-11-22
CI(translations), Documentation, Typing and an assortment of fixes

C0rn3j:
* Add compile_translations module to CI
* Fixes broken translation from #1224

* Added necessary definitions to phazor.c to pass MSVC compiler and move `unistd.h` under non-Windows block
  * I did not get Tauon to compile on MSVC as a whole in the end, but I am not hard stuck yet
* Added a bunch of Windows MSVC notes in `pyproject.toml` - untidy, for now

* Fix trying to use a hardcoded `/usr/share/TauonMusicBox` path even when it does not exist on the system
* run.sh fix for running from a downloaded zip file
* Renamed `.env` to the more standard `.venv`
* More spaces to tabs in main
* Fixed crashing on exit
* Removed unused `temp_audio` var
* Removed unused arg from `no_padding()`
* Converted more code to Path where applicable, converts to `str()` where necessary
* Changed `system` to use capitalized names, to match `platform.system()`
* Typing additions and fixes


Ricardo-Simoes:
How did my translations broke the translations???? I don't understand, please clarify.

C0rn3j:
> How did my translations broke the translations???? I don't understand, please clarify.

I've posted the log in https://github.com/Taiko2k/Tauon/pull/1224#issuecomment-2494709392, you can get it yourself by running `python compile_translations.py`.

There are now a couple duplicates.

It does look like the translations lack updates though, as "Broadcast This" does not even seem to be anywhere in the current code, I'll look into that.

C0rn3j:
@Ricardo-Simoes There was only one that wasn't an obvious duplicate, I presume I deleted the correct one?

```bash
# KEPT: 
msgid "There are no tracks in this playlist to broadcast."
msgstr "Não há faixas nesta lista de reprodução para transmitir."

# REMOVED:
#~ msgid "There are no tracks in this playlist to broadcast."
#~ msgstr "Não há faixas nesta playlist para transmitir."
```

PR #1311, opened 2024-11-23
PlayerCtl refactor, add TauonQueueItem class and start using TauonPlaylist too

C0rn3j:
This started as an attempt to migrate PlayerCtl into a separate file, which was a complete failure and I decided to instead type it completely and refactor things a little - which turns out needed to be a lot of refactoring, as playlists and queues, which I wanted typed, are used everywhere.

* Fully converted `t_main` to tabs, the entire Python codebase is now using tabs for indentation
* Fix f-strings to restore Python <3.12 support
* Typed up tagscan module
* Specify version import in `t_draw` too for Gtk - Drop all Gdk/Gtk libs down to 3.0 due to https://github.com/Taiko2k/Tauon/issues/1316
* `sg()` is now `get_track_in_playlist()`
* `g()` is now `get_track()`
* `QUE` is now `track_queue`
* `hide_title` is now bool instead of 0/1 int
* Added class `TauonQueueItem` and refactored `t_main` for it, also started using the earlier-added `TauonPlaylist` class after a refactor
  * DB bumped to v69 and a migration was added
  * The classes are converted to dicts during pickling/unpickling
  * This means that instead of guessing indexes we can just use the class properties:<br>
  ```python
                    # Skip if hide flag is set
                    if tab[8] is True:

                    # vs
  
                    if tab.hidden:
  ```
* Pickling now defines the highest available protocol version, it defaults to lowest available (currently 4 vs 5)
* Fixed wrong fallback URL for https://gensokyoradio.net
* Made some exceptions handle better, to not uselessly spam an entire traceback when we already know what's wrong
* Migrated `console.print()` to `logging` module in `PlayerCtl`
* PlayerCtl: Removed unused args
* PlayerCtl: Fix referring to an object of itself (`pctl`) instead of using a `self` reference - this is correct, right?
* PlayerCtl: Fix infinite requests timeout
* `tryint()` fix from earlier broke it completely, fix it for real this time, also moved it to Extra
* Config: Fix not being able to load negative int values and having them reset to default each time
* Phazor: Set libdir to lib64 instead of lib which is 32-bit dir on some distros
* Add devel stubs for Gtk3 and Gdk3, since that's about the only reasonable way we're getting them typed up when developing

TODO:  
* Consider if anything in `TauonQueueItem` or `TauonPlaylist` should not be changed, so we don't have to write another migration later
* Maybe we should be using slots for this - https://stackoverflow.com/questions/472000/usage-of-slots


Taiko2k:
In my experience pickeling classes is a bad idea. Better to save the dict `myclass.__dict__` then when restoring re instance a new class and use`__dict__.update(dict)`

C0rn3j:
Would welcome input on the last remaining TODO items. I suppose the slots are the least important item as we can always easily add them later without a migration(?).

Currently, we have `TauonPlaylist`, `TrackClass` and `TauonQueueItem`.

`TrackClass` would probably be best to renamed to `TauonTrack`.

I particularly do not like TauonPlaylist having a playlist_ids (renamed from the previous `playlist`). Playlist having playlist is confusing.

PR #1330, opened 2024-12-03
Cleanup and stuff

RSKYS:


Taiko2k:
I still planned on keeping backward compatibility with the old way for now.

RSKYS:
> I still planned on keeping backward compatibility with the old way for now.

Any particular reason why they're not in one anymore?

C0rn3j:
macOS support from what I've seen, you can merge them together and add a flag like --nopipewire.

Ideally, we could put compile translations, update translations and the two phazor compile scripts into one Python file.

Taiko2k:
If I'm itterating it's nice to select a specific script since it's faster to compile.

Taiko2k:
We cant remove the Windows build script unless we have a replacement

RSKYS:
See latest change.

RSKYS:
Commented out hard_fix for other time but don't touch it for now; I'll commit other stuff on phazor later as I'm guessing you're at the middle of something.

LGTM

run.sh
```diff
@@ -1,37 +1,120 @@
...
+    gcc src/phazor/kissfft/kiss_fftr.c src/phazor/kissfft/kiss_fft.c src/phazor/phazor.c \
+        $(pkg-config --cflags --libs python3 samplerate wavpack opusfile vorbisfile libmpg123 flac libopenmpt libgme libpipewire-0.3) \
+        -shared -o libphazor-pw.so -fPIC -Wall -O3 -g -DPIPE
+
+    mkdir -p build
+    mv libphazor-pw.so build/libphazor-pw.so
+}
+
+# Display menu
+show_menu() {
```
C0rn3j, line 97: You're reimplementing select, just use it instead

```bash
	select yn in "Yes" "No"; do
		case $yn in
			Yes ) answerEncrypt="yes"; break;;
			No ) answerEncrypt="no"; break;;
		esac
	done

```

run.sh
```diff
@@ -1,37 +1,120 @@
...
+    python -m venv .venv
+    source .venv/bin/activate
+    pip install -r requirements.txt requirements_optional.txt requirements_devel.txt build
+    python -m compile_translations
+    python -m build --wheel
+    pip install --prefix ".venv" dist/*.whl --force-reinstall
+}
+
+
+pw_pulse() {
```
C0rn3j: pw_pulse and pw_raw does not seem too explanatory.

Something like `compile_phazor` and `compile_phazor_pipewire` as it was sounds more readable to me and more accurate, as we're not necessarily supporting either pulse or pw backend on macOS/Windows. 

pw_pulse is also confusing a bit more as it does not contain pw support on any OS.

run.sh
```diff
@@ -1,37 +1,120 @@
...
-pip install -r requirements_devel.txt
-#pip install -r requirements_windows.txt
-pip install build
-python -m compile_translations
-python -m build --wheel
-#python -m installer --destdir=".venv" dist/*.whl
-pip install --prefix ".venv" dist/*.whl --force-reinstall
-
-tauonmb "$@"
+case $choose in
```
C0rn3j, line 112: Same here, just wrap it in select

run.sh
RSKYS: Oh that's actually good idea, but "pwr" and "pwp" are not necessary to get included in select as they're taiko thingy's (I mention them at invalid as (dev) anyway).
Also I needed to use -e not regular for echo output, that needs another push.
I think it's already readable and easy enough for future script addups too.

run.sh
RSKYS: Like first one.

run.sh
RSKYS: Fair enough, fixed.

C0rn3j:
This broke run.sh, and less importantly converted the whitespace back to spaces.

I'll PR a fix in a bit.

PR #1339, opened 2024-12-07
Fix incorrectly encoded XSPF tags

### LINKED ISSUE #1331: XSPF export produces incorrectly encoded tags
Commit 0a59adc04af6813d1bae9d737754af8d2ecfd98c fixed the issue https://github.com/Taiko2k/Tauon/issues/1088 , however, XSPF tags like `<title>`, `<creator>`, and `<album>` should not be encoded by `urllib.parse.quote`. Only content of the _`<location>`_ tag should be encoded that way to comply with the Spec.

iamalexei:
Fixes #1331

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -17070,23 +17070,17 @@ def export_xspf(pl: int, direc=None, relative=False, show: bool = True) -> int |
 
 		xspf_track_tag = ET.SubElement(xspf_tracklist_tag, "track")
 		if track.title != "":
-			xspf_title_tag = ET.SubElement(xspf_track_tag, "title")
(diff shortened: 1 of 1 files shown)


C0rn3j:
I think assigning to variables which are never used later is confusing and should be instead just done inline - presuming it works fine this way.

i.e. 
```python
if track.artist != "":
	ET.SubElement(xspf_track_tag, "creator").text = track.artist
```

instead of

```python
if track.artist != "":
	xspf_creator_tag = ET.SubElement(xspf_track_tag, "creator")
	xspf_creator_tag.text = track.artist
```

iamalexei:
Done. Agree, `ET.SubElement(xspf_track_tag, "creator").text = track.artist` looks cleaner

PR #1361, opened 2024-12-24
CI, phazor and other misc improvements and bug fixes

### LINKED ISSUE #616: Cover art for Jellyfin
Is there a way to show the art when streaming from Jellyfin?
I can't see them and if i try to quick-fetch it says "Cannot download art for network tracks".

### LINKED ISSUE #1244: Tauon does not attempt to recreate a sink/link(?) when the connection dies due to sound server restart or device removal
As per title.

I can have Tauon running and playing music:
image

Then reconnect some device, in this case I was using BT headset, ran `systemctl --user try-restart pipewire pipewire-pulse` and reconnected it:

image

And Tauon will now be stuck at 00:00 or whatever seek was, not playing anything, until I completely restart it.

It should recreate the connection if it dies.

Happens on Arch Linux with Tauon from AUR.

### LINKED ISSUE #1347: Shutdown crashes Tauon
Not really a huge problem, but it is unsightly.

```python
2024-12-13 15:56:38 [ INFO  ] [  t_main  ] Bye!
2024-12-13 15:56:38 [WARNING] [ warnings ] /usr/lib/python3.12/asyncio/base_events.py:726: ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
  _warn(f"unclosed event loop {self!r}", ResourceWarning, source=self)
 (warnings.py:112)
  File "/home/c0rn3j/Projects/Tauon/.venv/bin/tauonmb", line 5, in <module>
    from tauon.__main__ import main
ImportError: cannot import name 'main' from 'tauon.__main__' (/home/c0rn3j/Projects/Tauon/.venv/lib/python3.12/site-packages/tauon/__main__.py)
```

Code responsible should be at the bottom of https://github.com/Taiko2k/Tauon/blob/master/src/tauon/__main__.py

Does not seem to affect pyinstaller builds.

### LINKED ISSUE #1351: Change for building on linux
Hi, i've seen that the main script has been change on https://github.com/Taiko2k/Tauon/wiki/Building ref https://github.com/Taiko2k/Tauon/wiki/Building/_compare/18edb1de1c26ee286684a5c7178f8aa71931a3c1...038cb7f2c040bc47e8bc3e26a4018ed634ccc601

If i try to launch `python __main__.py` from the top folder it can't work, for sure `__main__.py` is not there, so if i try with `python src/tauon/__main__.py` it works but it create new `cache` & `user-data` folders in `src/tauon`.

Do i simply have to move my existing `cache` & `user-data` folders to `src/tauon` or is there a smarter way to handle this ?

### LINKED ISSUE #1358: ResourceWarning when reading album covers
```python
2024-12-23 16:31:25 [WARNING] [ warnings ] /usr/lib/python3.13/site-packages/tauon/t_modules/t_main.py:40921: ResourceWarning: unclosed file <_io.BufferedReader name='/home/user/Nextcloud/HugeFiles/Music/Mecha Maiko - Mad Soft Butt/cover.jxl'>                               
  album_art_gen.display(track, (x, y), (box, box)) (warnings.py:110)
```
```python
2024-12-23 16:31:25 [WARNING] [ warnings ] /usr/lib/python3.13/site-packages/tauon/t_modules/t_main.py:13720: ResourceWarning: unclosed file <_io.BufferedReader name='/home/user/Nextcloud/HugeFiles/Music/Mecha Maiko - Mad Soft Butt/cover.jxl'>                               
  self.im = album_art_gen.get_blur_im(track) (warnings.py:110)
```

Looks like there's even two of them

EDIT: there's considerably more I tagged here https://github.com/Taiko2k/Tauon/pull/1361/commits/7fbcf68ae55c2fd8051f3a90475cb1baaf5362d4

There's also these asyncio ones which I don't know exactly how are triggered:

```python
17:24:48 [ INFO  ] Done writing database
17:24:49 [DEBUG] Starting PHAzOR backend…
17:24:49 [DEBUG] Lib phazor found in C:\Users\Yeet\Desktop\TauonMusicBox-windows (3)\TauonMusicBox\dist\TauonMusicBox\_internal\phazor.cp312-mingw_x86_64_msvcrt_gnu.pyd
17:24:49 [ INFO  ] Open - requested start was 0 (0)
17:24:49 [ INFO  ] Extension: mp3
17:24:49 [ INFO  ] Transition jump
17:24:49 [WARNING] asyncio/base_events.py:725: ResourceWarning: unclosed event loop <ProactorEventLoop running=False closed=False debug=False>
```

Some of these are fixed by #1361

### LINKED ISSUE #1367: Rsvg deprecation
2024-12-30 22:37:33 [WARNING] [ warnings ] /home/c0rn3j/Projects/Tauon/.venv/lib/python3.13/site-packages/tauon/t_modules/t_svgout.py:60: DeprecationWarning: Rsvg.Handle.render_cairo is deprecated
  svg.render_cairo(svg_context)

```python
		viewport = Rsvg.Rectangle()
		viewport.x = 0
		viewport.y = 0
		viewport.width = width
		viewport.height = height
		svg.render_document(svg_context, viewport)
#		svg.render_cairo(svg_context)
```

This gets it going but it's obviously wrong, it renders, but looks way off.

EDIT: Fixed, PR'd.

C0rn3j:
Fixes made while attempting to deal with #1244 and some more.

is `prefs.cache_list` unused and safe to be removed?  
it looks to be called once from a module to init a cache list, and loaded from save[170], but it is never set.

* Make `spot_ctl`, `tidal`, `gall_ren` and `thumb_tracks` always belong to Tauon class and stop duping them globally too, this enables us to set them directly without having to check for None types
  * The code (both pre and post PR) looks as if Tidal is a hard dependency, was this intended, or am I reading it wrong?
* Remove `want_reconnect` in phazor - it did nothing.
* Start using `stdbool.h` in phazor and convert relevant ints to bool.
* Make phazor require C17 at minimum(2018), and prep for requiring C23 - which just came out in 2024-10 so best give it some time - that one adds nice features like... `bool`
* Switch zip releases to 7z, they are natively supported in both Linux and Windows (macOS uses DMG)
* Stop hardcoding Python versions in spec files
* Add `TaskbarLib.idl` source
* Fix missing `runtimeobject.lib` dependency in SMTC for Release builds
* Grab fonts in Windows CI as per https://github.com/Taiko2k/Tauon/pull/931#issuecomment-1464880133
* Pack `TauonSMTC.dll` and `librespot.exe` and `TaskbarLib.tlb` and `fonts` into Windows installs
* ~~Start shipping python-magic, sets up groundwork for #1361 - this needs fixing as per https://github.com/pyinstaller/pyinstaller/issues/8883, so import is commented out for now.~~
* Fixes #1347
* Fixes #1358 - by using context managers and implementing support for them in tagscan module among other things -  albeit the mentioned asyncio issue is still there
* Delete this seemingly useless line in `t_main.py` - `pl_thumbnail = GallClass(save_out=False)`
* Convert [install|user|config]_directory to Path
* Since we now ship `locale` inside the install_directory, always set it to `locale_directory = install_directory / "locale"` no matter if we're using Flatpak or are perhaps located in /opt
* Add some missing dependencies to myspac.txt for building manually on Windows
* Switch Windows CI to myspac.txt instead of duplicating the dependency list there
* Fix race condition with caching affecting Windows
* Fixed up the unused GuitarChords module a bit
* Modularize Prefs class
* Pack canberra-gtk-module to the Linux image
* Fixes #1367
* Made LoadImageAsset, WhiteModImageAsset and asset_loader() usable outside of t_main by properly defining their variables. For later rearranging. The function was converted to positional-only to prevent being able to mess it up, can be removed later as it is quite verbose this way.
* Fix minimode crashing when song isn't playing - minimode looks fairly blank if it's inited without a song, which is better than a crash
* Remove libwayland*.so files from Linux pyinstaller image due to https://github.com/pyinstaller/pyinstaller/issues/8963 - with a PR here https://github.com/pyinstaller/pyinstaller/pull/8964
* Fix Windows and Linux CI builds having useless nested directories
* Fix radio playback
* Fixes #616
* Fix crash when trying to copy current song title to clipboard when nothing is playing
* Fix user-data in case of running from a cloned directory to be in the root of the project as it was before. Also stop duping user_directory by using holder class - fixes #1351


Taiko2k:
I'd prefer to not add `python-magic`. I'm not convinced its necessary and would prefer to avoid adding more dependencies, especially non Python native ones. 

I don't really think importing files with incorrect extensions is an issue, that's on the user, and may even be that the user doesn't want those imported into a music player if they are named that way. Regardless this can be done without a library.

Taiko2k:
How about `python-puremagic`? At least that's more cross platform.

As for how I think this should be implemented, the type detection should go in the tag scan stage, so, if a file was detected incorrectly based on file ext during initial import, the tag scan stage should correct it. (then log a warning) (for speed, the import stage should avoid reading files)

Another opinion, I think this would be too minor of a feature to make optional, if it is to be implemented it should be required, regardless of if its programmatically optional.

C0rn3j:
> How about python-puremagic? At least that's more cross platform.

It looked to me like python-magic works on all platforms (Linux/macOS/Windows).

I will check out puremagic before implementing it. 

I've commented it out entirely in this PR, we can hash out the details about that one in the associated issue/Discord.

> As for how I think this should be implemented, the type detection should go in the tag scan stage, so, if a file was detected incorrectly based on file ext during initial import, the tag scan stage should correct it. (then log a warning) (for speed, the import stage should avoid reading files)

Sounds good to me

C0rn3j:
And I think that fixes the user dirs.

Portable cloned:  
image

Portable:  
image

Installed (hacked venv file for demo, hence it says portable):  
image

Taiko2k:
> is prefs.cache_list unused and safe to be removed?

Yeah it can be removed if its not used

> looks as if Tidal is a hard dependency, was this intended

Its currently hard, but probably should be optional, mainly since `python-tidalapi` is AUR and to make building easier I think its better to try make AUR modules optional.

PR #1384, opened 2025-01-08
Improve Maloja scrobbling

brunohpaiva:
Improvements for the Maloja scrobbling feature.

- Sending JSON data as recommended by the [API documentation](https://github.com/krateng/maloja/blob/master/API.md#api-documentation).
- Now using the `artists` field. The old `artist` field is [still supported](https://github.com/krateng/maloja/blob/c6cf28896ca836407a1943cae5542d2b83d009cf/maloja/apis/native_v1.py#L590), but it is likely deprecated as it is not mentioned in the [documentation](https://github.com/krateng/maloja/blob/master/API.md#submitting-a-scrobble).
- Added `album`, `albumartists`, `length`, and `time` fields.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -7592,7 +7592,7 @@ def maloja_get_scrobble_counts():
 	tauon.bg_save()
 
 
-def maloja_scrobble(track: TrackClass) -> bool | None:
+def maloja_scrobble(track: TrackClass, timestamp: int | None = None) -> bool | None:
 	url = prefs.maloja_url
 
 	if not track.artist or not track.title:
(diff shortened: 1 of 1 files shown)


src/tauon/t_modules/t_main.py
```diff
@@ -7603,13 +7603,24 @@ def maloja_scrobble(track: TrackClass) -> bool | None:
...
 
+	if timestamp is None:
+		timestamp = int(time.time())
+
 	d = {}
-	d["artist"] = track.artist
+	d["artists"] = [track.artist] # let Maloja parse/fix artists
 	d["title"] = track.title
+
+	if len(track.album) > 0:
```
C0rn3j, line 7613: You should be able to just do `if track.album:`, empty strings won't qualify.

src/tauon/t_modules/t_main.py
```diff
@@ -7603,13 +7603,24 @@ def maloja_scrobble(track: TrackClass) -> bool | None:
...
+		d["album"] = track.album
+	if len(track.album_artist) > 0:
```
C0rn3j, line 7615: Same as above

src/tauon/t_modules/t_main.py
```diff
@@ -7603,13 +7603,24 @@ def maloja_scrobble(track: TrackClass) -> bool | None:
 			url += "/"
 		url += "apis/mlj_1/newscrobble"
 
+	if timestamp is None:
+		timestamp = int(time.time())
```
C0rn3j, line 7607: We should set this as a parameter in the function definition instead, since we're setting it as default anyway, removes the None case entirely too:

 `timestamp: int = int(time.time())`

C0rn3j:
LGTM, presuming the changes were tested

PR #1388, opened 2025-01-12
Refactor initialization, functions and classes

C0rn3j:
image

This is effectively a complete refactor of Tauon's Python side.

This introduces a new dataclass called `Bag` in `t_main.py`, created after `Prefs` (technically it could be merged into `Prefs`, but that is big enough and it's not a Preference but mostly an object holder), which partially absorbs the initial `Holder` from `__main__.py`, and later the `Tauon` class absorbs `Bag` and `Holder`.

Things were rearranged to allow this hierarchy, most classes are now initialized within `Tauon` or `PlayerCtl`, as needed.

This allows us to pass `Holder`, `Bag` and later `Tauon` everywhere, allowing access to all the things that used to be either explicit or implicit globals, all the while allowing type checkers to understand the code and not crash/timeout as is the current case with `master`.

Works mostly fine, but the code is a little ugly at the moment in parts.

~~I've been chopping this PR up into pieces and sending the working bits to `master` in a reviewable fashion for now, as it touches 37K lines at the time of writing, out of the total 47K\~.~~

PR #1500 merged vast majority of the changes that this PR initially had.

What's left in this PR is timers, some init reworks and wrapping the code into a `main()` function.

---

Biggest problems with this WIP branch are:
  - [x] Some imports not working due to being imported in main() and not passed in any way
  - [x] Some random issues marked with a `TODO(Martin)`

---

The goals here are:
- [ ]  Stop using class globals - `LoadImageAsset`, `WhiteModImageAsset`, `TextBox` and `TextBox2` are left
- [x]  Rework functions and classes to be initialized without implicitly relying on globals
- [x]  Remove all explicit globals
- [x] ~~Convert Prefs to dataclass~~
- [x]  Get rid of `t_bootstrap.py` by turning its contents into a regular class object, currently it functions as a global object import of sorts. Instead we'll create a `main()` function in `t_main.py` to be able to pass it, to do so we'll have to get rid of doing everything in the top level of the module

---

2025-03-03 - `4352` Ruff issues, `27804` Pyright issues in master

2025-03-02 - `3968` Ruff issues, `14947` Pyright issues in PR
2025-03-03 - `2946` Ruff issues, `14261` Pyright issues in PR (1K Ruff issues are line-too-long which are hidden for now)
2025-03-04 - `2544` Ruff issues, `14468` Pyright issues in PR
2025-04-12 - `2552` Ruff issues, `13453` Pyright issues in PR
2025-06-10 - `2456` Ruff issues, `11965` Pyright issues in PR

### DIFF
--- a/src/tauon/__main__.py
+++ b/src/tauon/__main__.py
@@ -32,7 +32,7 @@ from gi.repository import GLib
 
 from tauon.t_modules.t_logging import CustomLoggingFormatter, LogHistoryHandler
 
-from tauon.t_modules import t_bootstrap
+from tauon.t_modules.t_bootstrap import Holder
 
 
 log = LogHistoryHandler()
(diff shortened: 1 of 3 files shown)


C0rn3j:
This is now finally ready.

Should be only pulled after #1621 as it'll need to be (automatically?) rebased against it, since it includes its commit.

Issue #1437, opened 2025-02-13
SDL3 issues


C0rn3j:
t_draw uses `SDL_RenderCopy`, `SDL_RenderCopyEx`, `SDL_SetColorKey`, `SDL_QueryTexture` and misuses `SDL_MapRGB`

---

<details><summary>Already fixed things:</summary>

Fixed in #1487 for Windows, #1488 for macOS and #1574 for Linux - We need to ship `metadata.json` or use specific env vars for PySDL3 to not have it download all extra SDL libraries and use our internal ones: https://github.com/Aermoss/PySDL3/issues/15#event-16303654791

Fixed in #1507 - #1491 transparent synced lyrics 

Fixed in #1506 - AttributeError: module 'sdl3' has no attribute 'SDL_TRUE' (also SDL_FALSE)

Fixed in #1479 - xwayland seems to be always used unless we force Wayland - `SDL_VIDEODRIVER=wayland tauonmb` - possibly due to the reasons listed in https://github.com/Taiko2k/Tauon/pull/1332#issuecomment-2523332876

Fixed in #1457 - `paste_lyrics()`, `paste_lyrics_deco()` and `standard_size()` miss sdl3 prefix

Fixed in #1457 - `t_draw` has:

```python
		wop = rw_from_object(g)
		self.surface = IMG_Load_RW(wop, 0)
```

Fixed in https://github.com/Taiko2k/Tauon/commit/4e6a27644b886049fb9deb9fb387989b6ecf9f74 - NameError: name `rw_from_object` is not defined

Fixed in https://github.com/Taiko2k/Tauon/commit/0be91c6929e5c08072731a0023ff015c8a70e79b - `SDL_WINDOW_FULLSCREEN_DESKTOP` is not a known attribute of module "sdl3"

Fixed in https://github.com/Taiko2k/Tauon/commit/e5422a7b8bc91a550623808394b704ae1a9dffe0 - module 'sdl3' has no attribute 'SDL_GL_GetDrawableSize'

Fixed in https://github.com/Taiko2k/Tauon/commit/71d2dd23e557b3f706856dcae4e934e60005faf7 - Can't launch on a Windows VM anymore, unsure if it works outside of a VM:

EDIT: It does work outside, is this a PySDL3 limitation since it requires to load everything and the kitchen sink?  
EDIT2: Tauon now explicitly selects OpenGL, we should try different renderers to enable VMs without 3D acceleration I suppose.

```python
[...]

PR #1440, opened 2025-02-14
Fix override logging lines

C0rn3j:
Since they're Path now it fails to concat.


msmafra:
https://github.com/Taiko2k/Tauon/issues/1437#issuecomment-2660169199

> [@msmafra](https://github.com/msmafra) Not related.
> 
> You can do `unset XDG_MUSIC_DIR` before launch for now, fixed by [#1440](https://github.com/Taiko2k/Tauon/pull/1440)

Actually I commented out all XDG dirs definitions for it to work. Did not test one by one, but commenting out only XDG_MUSIC_DIR did not do the trick.

C0rn3j:
XDG_DOWNLOAD_DIR had the same issue, rest should work

PR #1451, opened 2025-02-23
Improvements for phazor.c

RSKYS:


RSKYS:
Still testing things on this, marking it as draft so it won't get mistaken with ready.

src/phazor/phazor.c
```diff
@@ -567,8 +573,11 @@ int wave_error = 0;
 int16_t wave_16 = 0;
 
 int wave_open(char *filename) {
-
 	wave_file = fopen(filename, "rb");
+	if (wave_file == NULL) {
+		printf("pa: Error opening WAVE file\n");
```
jansol: It'd be helpful to also include the [reason why it failed](https://www.man7.org/linux/man-pages/man3/fopen.3.html#ERRORS):
```suggestion
		printf("pa: Error opening WAVE file: %s\n", strerror(errno));
```

RSKYS: Done, thanks for the suggestion :)

Issue #1464, opened 2025-02-27
No Album Covers in pre-releases


Crocodile73:
Not sure over reports that have been reported are exactly the same. No issue on 7.9.0

Its on the new preview build, i don't get any covers for albums at all. Running Windows 11 24H2

Albums on in D:\My Music then individual album folders, could it be because its an external drive?

Upload screenshot when i can

Crocodile73:
![Image](https://github.com/user-attachments/assets/dcb6b6a1-9e6d-47d4-a22f-187fa6e8bcca)

Crocodile73:
> Can you press ~ (or RMB -> Misc -> Toggle Console) and see what's in the log about it?

It says Gallery render error, when enabling it when library is already scanned.

First run log attached

Crocodile73:
[log.txt](https://github.com/user-attachments/files/19025560/log.txt)

C0rn3j:
> It says Gallery render error, when enabling it when library is already scanned.

Could you post the log from when that error happens?

Crocodile73:
[log.txt](https://github.com/user-attachments/files/19026326/log.txt)

C0rn3j:
Does the CI build from https://github.com/Taiko2k/Tauon/pull/1465 fix it?

https://github.com/Taiko2k/Tauon/wiki/Development-builds

Should be up in 15 minutes.

Crocodile73:
[log.txt](https://github.com/user-attachments/files/19028356/log.txt)

C0rn3j:
I don't get it.

```python

	def load_image(self, g: BytesIO) -> sdl3.LP_SDL_Surface:

		size = g.getbuffer().nbytes
		pointer = ctypes.c_void_p(ctypes.addressof(ctypes.c_char.from_buffer(g.getbuffer())))
		stream = sdl3.SDL_IOFromMem(pointer, size)

		return sdl3.IMG_Load_IO(stream, closeio=True)
```

The log would suggest we're calling `sdl3.IMG_Load_IO(stream, closeio=True)` with only one argument, but there's clearly two.

I cannot reproduce it either raw from master or from a pyinstaller build, but I can indeed reproduce it from the pyinstaller prerelease.

Ran into #1466 on top too.

[...]

PR #1467, opened 2025-02-28
Implement Inno Setup in CI

### LINKED ISSUE #1445: Include setup exe in release CI
Ive committed an Inno Setup file to `extra/setup.iss`.

 - We could use an existing action such as <https://github.com/marketplace/actions/inno-setup-action>.
 - Ive set a placeholder of `<???>` in paths in the script but these need to be set to the correct paths whatever they are.
 - We need to override the name of the output file to use our naming scheme e.g. ` Tauon.v7.9.0-1.x64.exe `.
 - Would be good to also override the app version number used in the script? Not sure how that can be done.

C0rn3j:
https://jrsoftware.org/ishelp/index.php

Fixes #1445

- [ ] Option flag - I left the default -O flag as per the example, is this okay?
- [ ] Dynamic version
- [x] *Warning: Architecture identifier "x64" is deprecated. Substituting "x64os", but note that "x64compatible" is preferred in most cases. See the "Architecture Identifiers" topic in help file for more information.* - https://jrsoftware.org/ishelp/index.php?topic=archidentifiers
- [x] *Warning: Constant "pf" has been renamed. Use "commonpf" instead or consider using its "auto" form.*
- [x] *Warning: Setting the [Setup] section directive "OutputBaseFileName" to "setup" is not recommended: all executables named "setup.exe" are shimmed by Windows application compatibility to load additional DLLs, such as version.dll. These DLLs are loaded unsafely by Windows and can be hijacked. Use a different name, for example "mysetup".*


C0rn3j:
I think this is ready for review & pull, the version atm has to be bumped in the CI .yaml which would be nice to automate later, but other than that it seems solid.

The build is somewhat broken due to #1464 but it's not related to the Inno steps.

PR #1473, opened 2025-03-02
Expanding search terms beyond

RSKYS:
Usually things coming back 10X times bigger from me; this was for `& <=> and` case and stuff in PR I canceled long ago.

These should make the dictionary expansion even easier for special character searches.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -34704,6 +34704,22 @@ def worker2():
 
 				o_text = search_over.search_text.text.lower().replace("-", "")
 
+				# Now users can add more special search terms to these.
+				n2_pairs = [
+					(" and", [" &"]), (" &", [" and"]),
+					(" to", [" 2"]), (" 2", [" to"]),
+					(" two", [" 2"]), (" 2", [" two"]),
+				]
+
+				n2_mode = False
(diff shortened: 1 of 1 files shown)


RSKYS:
Preview of the actual stuff:

image
image

src/tauon/t_modules/t_main.py
```diff
@@ -34704,6 +34704,22 @@ def worker2():
 
 				o_text = search_over.search_text.text.lower().replace("-", "")
 
+				# Now users can add more special search terms to these.
+				n2_pairs = [
+					(" and", [" &"]), (" &", [" and"]),
+					(" to", [" 2"]), (" 2", [" to"]),
+					(" two", [" 2"]), (" 2", [" two"]),
+				]
```
C0rn3j, line 34712: Why is the second tuple param a list?

Also is there any point in duplicating the pairs instead of defining it once and checking both combinations? 

i.e. `(" and", " &")` and simply checking both `and->&` and `&->and` below

RSKYS: Yeah, this was mostly the only way to make it work along extra space; and right now this just works, I can make it better later;

> Also is there any point in duplicating the pairs instead of defining it once and checking both combinations?

It does not work without overriding loop on it

PR #1350, opened 2024-12-18
Use Listening activity type in Discord RPC

brunohpaiva:
Discord added support for different activity types some time ago, but Tauon still uses the "Playing" activity type:

image image

`pypresence` got [support for it](https://github.com/qwertyquerty/pypresence/commit/f856ccaaeb2321f64f9692b75dc3ceda5c927f42) on August 2024, but the latest pypi release (4.3.0, from 2023) is outdated.

This PR implements it, fetching `pypresence` via Git.

image image

PR #1498, opened 2025-03-15
Rewriting rpc stuff, again?

RSKYS:
Another PR for https://github.com/Taiko2k/Tauon/pull/1350 afterward

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -34181,22 +34181,24 @@ def discord_loop() -> None:
 
 			title = _("Unknown Track")
 			tr = pctl.playing_object()
-			if tr.artist != "" and tr.title != "":
+			if tr.artist != "" and tr.title != "" and pctl.playing_state == 3:
 				title = tr.title + " | " + tr.artist
-				if len(title) > 150:
-					title = _("Unknown Track")
-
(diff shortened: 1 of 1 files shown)


C0rn3j:
You've managed to change the whitespace in this PR

RSKYS:
Is it incorrect? I did them all manually in vscode with ctrl [], thout this might be better

C0rn3j:
Yes, Tauon uses tabs, spaces are not better, they force your preferred indent-width onto others, meanwhile you can set tabs to your preference in your editor.

RSKYS:
Huh, cute
Never thought it'd be a thing, I'm sure I see spaces among codes too though time to time

Taiko2k:
~~What in the unholy abomination wh~~, ok I think we can just set these params to whatever the default is rather than this unpacking judo right? Like 

`**({"large_text": album} if album and not pctl.playing_state == 3 else {}),`

to

`large_text = album if album and not pctl.playing_state == 3 else None,`

though at some point it would be better to just be verbose and set these explicitly before the function call. And preferably with comments as to reasoning

```python
state = None
# If it is a radio station playing then we set state to album because yada yada
if pctl.playing_state == 3:
   state = album
   
RPC.update(state=state...
```

PR #1507, opened 2025-03-19
Fix synced lyrics panel drawing

### LINKED ISSUE #1491: Lyrics panel is transparent in Tracks + Art mode when song has synced lyrics
 - OS: Windows
 - OS version: 10
 - Tauon Version: 8.0.0
 - Installation method: Portable Binary

**Describe the bug**
Lyrics panel background is transparent when using the Tracks + Art mode. Screenshot attached. This only happens for synced lyrics. Normal lyrics do have the solid background.

**To Reproduce**
Open a song that has synced lyrics.

**Expected behavior**
Expected solid color in the lyrics panel rather than transparent.

**Screenshots/Videos**
If applicable, add screenshots/videos to help explain your problem.

Image

C0rn3j:
Fixes #1491, but I am unsure if I am putting things in the correct place.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -6433,7 +6433,7 @@ class TimedLyricsRen:
 			bg = (bg[0],bg[1], bg[2], 255)
 			font_size = 15
 			spacing = round(17 * gui.scale)
-			ddt.rect((window_size[0] - gui.rspw, gui.panelY, gui.rspw, window_size[1] - gui.panelY - gui.panelBY - l_panel_h,), panel_bg)
+			ddt.rect((window_size[0] - gui.rspw, gui.panelY, gui.rspw, window_size[1] - gui.panelY - gui.panelBY - l_panel_h), bg)
 		else:
 			bg = colours.playlist_panel_background
 			font_size = 17
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -6429,11 +6429,10 @@ class TimedLyricsRen:
 		highlight = True
 
 		if side_panel:
-			bg = colours.top_panel_background
+			bg = colours.side_panel_background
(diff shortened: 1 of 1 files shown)


src/tauon/t_modules/t_main.py
```diff
@@ -6432,6 +6432,9 @@ def render(self, index: int, x: int, y: int, side_panel: bool = False, w: int =
 			bg = colours.top_panel_background
 			font_size = 15
 			spacing = round(17 * gui.scale)
+			panel_bg = colours.side_panel_background
```
Taiko2k, line 6435: Rather than creating a new `panel_bg` variable should use the above `bg` variable as this is (often) used later on. 

Should be OK to set this to whatever here, `colours.side_panel_background` should be fine. I was using a different colour here for synced but this perhaps is more consistent.

src/tauon/t_modules/t_main.py
```diff
@@ -6432,6 +6432,9 @@ def render(self, index: int, x: int, y: int, side_panel: bool = False, w: int =
...
+			panel_bg = (panel_bg[0], panel_bg[1], panel_bg[2], 255)
+			ddt.rect((window_size[0] - gui.rspw, gui.panelY, gui.rspw, window_size[1] - gui.panelY - gui.panelBY - l_panel_h,), panel_bg)
```
Taiko2k, line 6437: Yes I think this should be here, but we do inherit `x,y,w,h` from the function parameters so we should use those.

C0rn3j: Those params set the lyrics a bit differently from the normal lyrics though so using straight out x,y,w,h without any edits does this:

image

src/tauon/t_modules/t_main.py
C0rn3j: switched it to side_panel_background

PR #1554, opened 2025-04-10
Enable fine grained control over desktop notifications

### LINKED ISSUE #1047: Generic notification on track change w/ KDE Plasma
Installed with Flatpak 
v7.6.7
on Kubuntu 23.04
w/ KDE Plasma 5.27.4 

When the application does not have focus and the track changes, it displays a generic notification under 'Plasma Workspace'. Is this supposed to happen? I would assume that it would have its own application slot so I could customise it's behaviour in KDE settings. I also would assume that the album art should be shown instead of a generic KDE logo. 

Currently I personally have Plasma Workspace notifications set to make a sound when they pop... obviously this is not the setting I want applied for music app notifications - so I have disable this effect globally just for Tauon. These notifications stack up and pollute the notification drawer unless I disable history for all Plasma Workspace notifications as well.

Love the application and greatly appreciate the amazing work you have done already!

_Example Notification_
image

_KDE Notification Settings_
image

andreyfedoseev:
Fixes #1047

DEs such as Gnome and KDE allow to configure desktop notifications per-application. To enable that, the following line should be present in the `.desktop` file:

```
X-GNOME-UsesNotifications=true
```

See https://developer.gnome.org/documentation/tutorials/notifications.html#disabling-notifications

### DIFF
--- a/extra/com.github.taiko2k.tauonmb.desktop
+++ b/extra/com.github.taiko2k.tauonmb.desktop
@@ -22,6 +22,7 @@ Terminal=false
 Type=Application
 Categories=AudioVideo;Player;Audio;
 Actions=PlayPause;Previous;Next;Stop
+X-GNOME-UsesNotifications=true
 
 [Desktop Action PlayPause]
 Exec=com.github.taiko2k.tauonmb.sh --no-start --play-pause
--- a/extra/tauonmb.desktop
+++ b/extra/tauonmb.desktop
@@ -22,6 +22,7 @@ Terminal=false
 Type=Application
 Categories=AudioVideo;Player;Audio;
 Actions=PlayPause;Previous;Next;Stop
+X-GNOME-UsesNotifications=true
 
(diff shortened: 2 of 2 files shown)


andreyfedoseev:
@C0rn3j 

I couldn't find any documentation on that, but it was mentioned in the [Release Notes for KDE Frameworks 5.57.0](https://kde.org/announcements/frameworks/5/5.57.0/#kservice)

I'm using KDE Plasma myself and I tested it locally by modifying the `.desktop` file. It is now displayed in the list of application in the notification settings.

image

andreyfedoseev:
> How do you use Tauon, Flatpak?

Yes, I use Flatpak.

It's possible that Plasma needs to reload the `.desktop` file after you make changes to it. There may be some clever way to do that, but what I do is just open the entry in the menu editor, add some dummy changes there and save it. That worked for me.

PR #1560, opened 2025-04-12
Make sure source_image is closed to prevent resource leaks

C0rn3j:
Fixes some image leaks when first loading an album.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -5348,8 +5348,9 @@ class ThumbTracks:
 				im.thumbnail((512, 512), Image.Resampling.LANCZOS)
 				width, height = im.size
 				data = im.tobytes()
+			source_image.close()
 			return GdkPixbuf.Pixbuf.new_from_data(data, GdkPixbuf.Colorspace.RGB, False, 8, width, height, width * 3)
-		except:
+		except Exception:
 			logging.exception("Error create pixbuf of album art")
 			return None
 
@@ -5368,13 +5369,12 @@ class ThumbTracks:
 			return t_path
 
 		source_image = self.album_art_gen.get_source_raw(0, 0, track, subsource=source)
-
 		with Image.open(source_image) as im:
 			if im.mode != "RGB":
 				im = im.convert("RGB")
 			im.thumbnail((1000, 1000), Image.Resampling.LANCZOS)
-
 			im.save(t_path, "JPEG")
+		source_image.close()
 
 		return t_path
 
@@ -20430,6 +20430,7 @@ class AlbumArt:
 		return pic
 
 	def get_source_raw(self, offset: int, sources: list[tuple[int, str]] | int, track: TrackClass, subsource: list[tuple[int, str]] | None = None):
+		"""Caller has to call .close() on the returned object afterwards"""
 		source_image = None
 
 		if subsource is None:


Taiko2k:
We need to encase these in `if not track.is_network`. Its a bit of a hack workaround since technically we don't have authority to close it in this case.

C0rn3j:
How can I test that this indeed ends up being an issue?

Because if that's indeed the case, `source_image` needs to be assigned differently, as it's currently only explicitly defined within the function and not used further, and not defined outside of it.

Taiko2k:
You’ll get an object is closed error when playing airsonic etc tracks. It’s because an underlying function caches this to RAM so the image doesn’t get re downloaded every access, but if source image gets closed then that breaks the cached version.

This is a hacky workaround but any proper solution would be work for what’s mostly a non issue.

C0rn3j:
Am sure that's the case for functions that re-use the var later, but the two functions I touched truly seem to not be affected by this, as they only use it temporarily?

path() saves the image `im.save(t_path, "JPEG")`, no longer needing `source_image`.

pixbuf() passes something from `data` forwards, not `source_image`.

I seem to be able to use Jellyfin just fine on this branch and resetting image cache repeatedly/restarting Tuaon, although am not 100% I am hitting the correct codepaths.

Taiko2k:
The issue is that if the function or other functions are called again, it might use cached source image.

If the source is a file on disk then it’s not a big issue to open it again, but once you call close on data stored in memory the data is gone.

There’s probably better ways this image system could be designed, but the immediate effect of this PR introduces this possible invalid state.

C0rn3j:
I have checked the code again and I don't see how this could happen, I've tested that it runs too.

`source_image` gets generated anew every time by the underlying functions and never used again as it is not a global, it just dies?


```python
	def get_source_raw(self, offset: int, sources: list[tuple[int, str]] | int, track: TrackClass, subsource: list[tuple[int, str]] | None = None):
		source_image = None

		if subsource is None:
			subsource = sources[offset]

		if subsource[0] == 1:
			pic = self.get_embed(track)
			assert pic
			# New BytesIO
			source_image = io.BytesIO(pic)
		elif subsource[0] == 2:
			try:
				if track.file_ext in ("RADIO", "Spotify") and self.pctl.radio_image_bin:
					return self.pctl.radio_image_bin

				cached_path = os.path.join(self.tauon.n_cache_directory, hashlib.md5(track.art_url_key.encode()).hexdigest()[:12])
				if os.path.isfile(cached_path):
					# New BytesIO
					source_image = open(cached_path, "rb")
				else:
					if track.file_ext == "SUB":
						# New BytesIO
						source_image = self.tauon.subsonic.get_cover(track)
					elif track.file_ext == "JELY":
						# New BytesIO
						source_image = self.tauon.jellyfin.get_cover(track)
					else:
						response = urllib.request.urlopen(self.tauon.get_network_thumbnail_url(track), context=self.tls_context)
						# New BytesIO
						source_image = io.BytesIO(response.read())
					if source_image:
						with Path(cached_path).open("wb") as file:
							file.write(source_image.read())
						source_image.seek(0)
			except Exception:
				logging.exception("Failed to get source")
		else:
			# New wrapper
			source_image = open(subsource[1], "rb")

		return source_image


	def get_cover(self, track_object: TrackClass):
		response = self.r("getCoverArt", p={"id": track_object.art_url_key}, binary=True)
		# We create new BytesIO here
		return io.BytesIO(response)
```

Taiko2k:
Oh right that's `get_source_raw`, I was checking the wrong function, yeah that should be fine

PR #1580, opened 2025-04-18
Deal with multiple albumartist tags in FLAC and OPUS files

### LINKED ISSUE #1578: Use all or first album artist value if there's more than one
When a track has multiple album artist values in its metadata (e.g., Artist A; Artist B), Tauon displays only the last value (Artist B). Other players like Spotify, foobar2000, or MusicBee either display all artists or the first artist for consistency.

This affects the user experience, especially for compilations or collaborations, where the displayed artist may not be the primary one, leading to confusion in navigation, sorting and scrobbling.

<details>

<summary>
Example screenshot
</summary>

Image

</details>

C0rn3j:
Fixes #1578

Looks like `albumartists` can be defined multiple times too, so account for it instead of throwing everything but the last definition.

It's not ideal, as we save the first album artist only, but current behavior only uses the last one, whereas the first one is seemingly always the actual main album artist, later ones being secondary.

Second commit simplifies the module a bit by means of introducing a base class.

### DIFF
--- a/src/tauon/t_modules/t_extra.py
+++ b/src/tauon/t_modules/t_extra.py
@@ -44,6 +44,7 @@ if TYPE_CHECKING:
 	from collections.abc import Callable
 
 	from tauon.t_modules.t_main import TrackClass
+	from tauon.t_modules.t_tagscan import TrackFile
 
 @dataclass
 class RadioStation:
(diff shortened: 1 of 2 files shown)


Taiko2k:
> self.sample_rate = 0   # TODO(Martin): Do we need to set this to zero, isn't default 48k fine?

`sample_rate` here is for informational purposes only. If the sample rate is not known then this value should be `0` as an alias for "unknown".

PR #1598, opened 2025-04-29
Ffmpeg and streaming changes

C0rn3j:
Bump downloaded Windows ffmpeg to 7.1.1 (from 5.0.1).

Stop trying to find ffmpeg/ffprobe binaries each time they're needed, cache them instead.

Start checking if ffmpeg/ffprobe was actually found before attempting to use the result - this might have been logically impossible even in the current code, but the functions themselves had no checks and it made Pyright angry.

Some small typing changes.


C0rn3j:
I don't want to deal with the implication that the ffmpeg binary gets yeeted out during runtime and it's not that much of a useful change, I'm closing this one.

PR #1626, opened 2025-05-19
Add language redirection for last.fm bio and Wikipedia

Lorsoen:
Artist bio (Ctrl + O) and "Search artist on Wikipedia" were always redirecting to English before.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -3887,7 +3887,7 @@ class LastFMapi:
 				l_artist = pylast.Artist(
 					artist.replace("/", "").replace("\\", "").replace(" & ", " and ").replace("&", " "),
 					self.lastfm_network)
-				bio = l_artist.get_bio_content()
+				bio = l_artist.get_bio_content(language=var_lang)
 				# cover_link = l_artist.get_cover_image()
 				mbid = l_artist.get_mbid()
 				url = l_artist.get_url()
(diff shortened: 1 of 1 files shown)


src/tauon/t_modules/t_main.py
```diff
@@ -12858,7 +12858,8 @@ def ser_gen(self, track_id: int, get_lyrics: bool = False) -> None:
 	def ser_wiki(self, index: int) -> None:
 		if len(self.pctl.master_library[index].artist) < 2:
 			return
-		line = "https://en.wikipedia.org/wiki/Special:Search?search=" + urllib.parse.quote(self.pctl.master_library[index].artist)
+
+		line = "https://" + lang_var + ".wikipedia.org/wiki/Special:Search?search=" + urllib.parse.quote(self.pctl.master_library[index].artist)
```
C0rn3j, line 12862: Would be best to convert it to an f-string while you're at it.

i.e prepend quotes with `f` and wrap all variables within `{}`

Lorsoen: cbd7d047

Taiko2k:
Language vars don't directly convert like that. It could be `en_US` or even a list such as `de_DE.UTF-8:fr_FR:en_US`, you need to extract the primary 2 letter language code. (And even then I'm not sure that covers all mappings, but it could be a start)

Also this `lang_var` variable needs to be validated, it must not be empty or None if its going to be used in a URL.

PR #1655, opened 2025-06-13
Some theming/visual changes

rexendevar:
1. New optional, themeable colors:
- "lyrics text" or "lyrics showcase" (`colours.lyrics`) - already existed but didn't seem to do anything before?
- "active lyric" (`colours.active_lyric`)
- "lyrics panel" (`colours.lyrics_panel_background`) - does not apply to showcase mode. couldn't figure out how. 
- "status text" (`status_text_over`, determines `status_text_normal`) - didn't realize these were already in the spec at the time. but it looked like they would have been overwritten anyway if "top panel" were defined afterwards.

All of these are handled gracefully if a theme does not define them.

2. When showing unsynced lyrics, attempt to visually remove LRC data if it exists. 

Apologies for some of the jank in here. I'm just learning Python lol u understand


rexendevar:
oh my god i didn't realize it's gonna be all 50 commits. also here's an example theme i made
[Sunken Translucent.txt](https://github.com/user-attachments/files/20733905/Sunken.Translucent.txt)

for posterity: the reason there's so many commits is because, due to dependency issues, i am unable to build the program on my computer, or run it until it is built. so every change i make has to be committed and then built by the github actions in order for me to test it.

src/tauon/t_modules/t_main.py
```diff
@@ -14,7 +14,7 @@
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
-#
+#
```
C0rn3j: end of lines shouldn't have whitespace

src/tauon/t_modules/t_main.py
```diff
@@ -19231,12 +19253,13 @@ def __init__(self) -> None:
 
 	def get(self, track: TrackClass) -> str:
 		if track.lyrics:
-			return track.lyrics
-		if track.is_network:
+			data = track.lyrics
+		elif track.is_network:
 			return ""
-		if track == self.cache_key:
+		elif track == self.cache_key:
```
C0rn3j: These were plain IFs because the previous code block `return`s, if it returns, there's no else to worry about, so no need to have extra else logic/indentation levels there.

I suggest configuring your code editor to be able to use the rules in pyproject.toml, as Ruff would highlight this for you

src/tauon/t_modules/t_themeload.py
```diff
@@ -236,8 +247,14 @@ def load_theme(colours: ColoursClass, path: Path) -> None:
 				colours.menu_highlight_background = get_colour_from_line(p)
 			if "menu border" in p:
 				colours.menu_tab = get_colour_from_line(p)
-			if "lyrics showcase" in p:
+				
+			if "lyrics showcase" in p or "lyrics text" in p:
 				colours.lyrics = get_colour_from_line(p)
+				lyrics_text_color_defined = 1
+			if "active lyric" in p:
+				colours.active_lyric = get_colour_from_line(p)
```
C0rn3j, line 248: Class shouldn't get variables defined that are not initialized there, so it should be added to ColourClass.

i.e. if you defined it like this:

`self.menu_background: ColourRGBA | None = None`

You can then also check if the color is set by checking the variable, without having to have extra flags for it.

C0rn3j:
> oh my god i didn't realize it's gonna be all 50 commits

It's not a huge issue, it can be squashed when being merged.

Though I highly suggest you get comfy with git, to be able to edit the history.  
Grab the hash of your earliest commit, `git rebase -i 5894a241fd84c0dc5c77504187a7ed47c05e19c4~1` (replace with your hash), and then you can easily do things like squashing the commits.  
Then force push it after you do the rebase and you double check it looks right.

src/tauon/t_modules/t_themeload.py
rexendevar: I don't think I understand this one. I set defaults for these colors in the __init__ for ColoursClass, so it's not useful to check whether they *exist* to find out if they have been *specifically set by a theme*.

The way I've done it, there's a path for if no theme runs at all (the defaults in colours.__init__), a path for an older theme running (the fallbacks - some colors have to be generated based on other colors that are stated explicitly), and a path for a new theme (new colors are set explicitly and the fallbacks should be bypassed).

Though now that I think about it, neither lyric text and active lyric are actually generated from other colors, so I don't need a fallback for them outside of ColoursClass.__init__. I'll remove those two - please let me know what I should do about the others

rexendevar: Just realized that themeload actually runs on startup, so there should never be a situation where no theme is loaded at all. But the new colors are still generated by the theme, so the best I can do is define them as None in ColoursClass.\_\_init__ and put the fallback bits in post_config()? does that sound like the right move?

rexendevar: Take a look at it now, I think this is all within spec. Thank you for your patience

rexendevar (commented):
I think it's all done and good please take a look love you

src/tauon/t_modules/t_themeload.py
C0rn3j: I've just quickly skimmed the changes to be fair, the important bit was making sure the variables are defined in ColoursClass and not tacked-on adhoc, if you do need a variable to track if a change was made, that's all good, was just suggesting a way to do it without it.

Although it may or may not make more sense to always track it via an extra var in that logic, especially since the default may change later.

Taiko2k:
Could you test the theme `lavender light`, and confirm the lyrics look good under that theme. Add a `lyric` colour definition to that file if needed.

rexendevar:
wow, it looks hideous. i'll add one hold on

rexendevar:
i didn't touch any of the light mode stuff making this patch cause i'm not a light mode kinda guy. but light mode definiitely will mess with a couple things.

rexendevar:
if i'm honest i think the fact that light mode is part of the theme file structure itself, rather than a user facing option, makes it no longer very useful now that basically every color can be explicitly themed. i would recommend phasing it out in favor of more explicit themed colors but that's beyond the scope of this pull

PR #1659, opened 2025-06-14
Fix Left align title style vanishing text

### LINKED ISSUE #1658: "Left align title style" causes album/folder text to vanish
Pop OS 22.04
Tauon development - occurs at least as far back as a week ago, does not occur in 8.0.1

**Describe the bug**
When "Left align title style" is checked, the relevant titles will go through a gradient for a moment before fully appearing black. I believe there might be something incrementing the opacity of colours.folder_title downwards.

**Screenshots/Videos**
The following image is what occurs for one frame. On the next frame it refreshes and you can't see the text at all
Image

colour = colours.folder_title
colour.a = max(colour.a - 50, 0)
^ probably the lines that are broken

rexendevar:
Fixes #1658

probably there's a better syntax for this but i wanted to make sure i wasn't using a pointer 

---

EDIT(Martin): Fixed up the syntax to tag the relevant issue, title and used deepcopy.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -29732,7 +29732,10 @@ class StandardPlaylist:
 					was = False
 					run = 0
 					duration = get_display_time(total_time)
-					colour = colours.folder_title
+					colour = ColourRGBA(colours.folder_title.r,
+							    colours.folder_title.g,
+							    colours.folder_title.b,
+							    colours.folder_title.a)
 					colour.a = max(colour.a - 50, 0)
 
 					if prefs.append_total_time and duration:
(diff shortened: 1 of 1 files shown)


rexendevar:
oh god hold on

C0rn3j:
You're taking a ColourRGBA object and creating a ColourRGBA object from all of its values, which means it's identical - it makes no sense that this would fix anything, are you sure?

EDIT: I suppose it actually does if the two variables need to be separate, as you're creating a *new* object here that won't be changed along with this one.

> probably there's a better syntax for this

We could edit ColourRGBA to add a copy constructor - https://stackoverflow.com/a/16046733

---

Another issue is that tabs are dynamic indentation, hence if you do this, it may look okay to you, but it will look broken to someone with a different tab width.

image

vs my setup

image


If you want to visually space variables, you need to make sure the tab amounts match, in this case the bottom three lines get two extra tabs, i.e. if you break off the first element onto a new line, the issue is gone and you don't need to use spaces for it in the first place.

PR #1681, opened 2025-06-20
Suggest user to make a bug report if crash log is detected

C0rn3j:


Taiko2k:
SDL has a native message box function. Let’s not add another GUI library.

C0rn3j:
I'll clean this PR up later:

image

EDIT: Ready to be merged

Issue #1625, opened 2025-05-19
Tauon fails to launch with system sdl3 on dev installs


farahnur42:
**Desktop (please complete the following information):**
 - OS: Arch on WSL
 - OS version: rolling
 - Tauon Version: commit 0faa1368
 - Installation method: built from source

**Describe the bug**
Tauon fails to launch with similar error messages as https://github.com/Taiko2k/Tauon/issues/1527

**To Reproduce**
1. Clone the repository
2. Use e.g. `uv sync` to set up the local environment
3. Run tauon via e.g. `uv run tauonmb`

**Expected behavior**
Tauon launches.

**Tauon log**
```log
❯ uv run tauonmb
2025-05-19 19:40:08 [ INFO  ] [ __main__ ] Tauon v8.1.0
2025-05-19 19:40:08 [ INFO  ] [ __main__ ] Copyright 2015-2025 Taiko2k captain.gxj@gmail.com

Error: Invoked an unimplemented function: 'SDL_SetHint'.
Error: Invoked an unimplemented function: 'SDL_SetHint'.
Error: Invoked an unimplemented function: 'SDL_SetHint'.
Error: Invoked an unimplemented function: 'SDL_SetHint'.
Error: Invoked an unimplemented function: 'SDL_SetHint'.
Error: Invoked an unimplemented function: 'SDL_Init'.
Error: Invoked an unimplemented function: 'SDL_GetError'.
Error: Invoked an unimplemented function: 'SDL_CreateWindow'.
2025-05-19 19:40:10 [ ERROR ] [ __main__ ] ERROR CREATING WINDOW! (__main__.py:334)
2025-05-19 19:40:10 [ ERROR ] [ __main__ ] Title: b'Tauon' (__main__.py:335)
2025-05-19 19:40:10 [ ERROR ] [ __main__ ] X: 536805376 (__main__.py:336)
2025-05-19 19:40:10 [ ERROR ] [ __main__ ] Y: 536805376 (__main__.py:337)
2025-05-19 19:40:10 [ ERROR ] [ __main__ ] Size 0: 1920 (__main__.py:338)
2025-05-19 19:40:10 [ ERROR ] [ __main__ ] Size 1: 1044 (__main__.py:339)
2025-05-19 19:40:10 [ ERROR ] [ __main__ ] Flags: 1073750064 (__main__.py:340)
Error: Invoked an unimplemented function: 'SDL_GetError'.
[...]

PR #1686, opened 2025-06-25
Add a workaround for #1625

C0rn3j:
Aims to workaround the bug in #1625 but PySDL3 has [another bug](https://github.com/Aermoss/PySDL3/issues/36) which prevents importing, so the PR is broken with current versions.


C0rn3j:
Latest PySDL3 should be able to do a reload correctly, merging.

Issue #1612, opened 2025-05-08
Two-way sync for playlist files. PLEASE


rexendevar:
**Is your feature request related to a problem? Please describe.**
I want to sync my playlists between my phone and my computer the same way I sync my audio files. I can sync the playlist files just fine, but if the program on either end refuses to take these files seriously, the whole exercise is pointless. Tauon does not take them seriously.

**Describe the solution you'd like**
I want Tauon to automatically scan, import and proactively update playlist files within the designated music folder, and I want it to immediately export all changes back to these same files without being asked.

**Describe alternatives you've considered**
Since quitting Spotify in 2023, I've tried Musicbee, Rhythmbox, Spotify again but for files, VLC, Jellyfin, Finamp, Amarok, Elisa, Strawberry, Museeks, Deadbeef and Audacious. All of em sucked for one reason or another. Tauon is the closest there is to perfection. 

**Additional context**
Here is the specific list of issues I have seen:

- Tauon does not automatically scan for playlist files. They must be manually imported every time.
- .m3u8 playlists will have a period added to their name on import.
- If a playlist file is created outside of Tauon while Tauon is running, it must be restarted or it will fail to import the playlist.
- Even after a playlist file is imported, Tauon will not notice changes in it. Not after rescanning the database, not after restarting the program, nothing. This alone makes two-way sync impossible.
- Even if this weren't true, every playlist that you want to sync *back to its file* must be manually set to export. If the original file was an .m3u8, Tauon will create a new file ending in ..m3u (notice the double period).
- I also don't understand what is required for Tauon to auto-export. I expect it to happen whenever the contents of the playlist are changed, but so far in testing I've had to manually re-export every time.

With all these problems, I would like to reiterate why this use case is still relevant, and why you should still care rather than give it up as an edge case. 

In this post-streaming world, I have gotten sick of Spotify and made the decision to switch back to file-based music exclusively. However, I don't want to sacrifice convenience; I use Syncthing to automatically sync all of this music from my computer to my phone and vice-versa whenever there are any changes. This works flawlessly - every song is accessible within seconds. 

[...]

PR #1688, opened 2025-06-26
Significantly improve playlist file handling

rexendevar:
My git workflow is filthy so you'll have to veto a lot of these changes. Like a lot of them. Sorry. I'm submitting this with the intention of fixing it over time, as it is fully feature complete for everything I need it to be. 

New features:
- Updated playlist import/export dialog. 
- Playlists now default to a specific filepath rather than simply a containing folder. (If the path ends in a trailing slash it will be treated like a folder though.)
- When a playlist specifies a direct filepath, Tauon can also automatically IMPORT changes from this file, if its filesize changes. This occurs when the user switches to the playlist.
- Newly imported playlists default to auto-export, auto-import and relative file paths.
- New "default playlist export folder" option in Function settings. This determines the folder that playlists will export to by default.
- Optionally, this folder can also be automatically scanned for new and unimported playlist files. This scanning will only occur when the user runs a manual database rescan.

Limitations/known issues/possibilities:
- What if you update a playlist file that should auto-import, then close Tauon without switching to it? Only just thought of this, have to handle it
- When a playlist file is rescanned, it won't do anything with newly found radio stations. I don't plan to change this unless others consider it important
- Windows/Mac compatibility definitely broken


Taiko2k:
The playlist import code really needs to be its own function, something like `pctl.check_update_playlist_from_file()` . Because we will want to run this elsewhere, maybe even on startup, not just switch playlist.

rexendevar:
> Because we will want to run this elsewhere, maybe even on startup, not just switch playlist.

How about we only run it if "auto regenerate playlists" is enabled?

I realize that a lot of Tauon's design is geared toward doing a bunch of stuff at the very start and end and not really touching the disk during runtime, but in my testing the biggest slowdown I experienced was MAYBE a quarter second, and that's on a playlist of 500 songs. To me that seems like a small price to pay for what ends up being a pretty big feature.

I'm assuming you mean to take it out of switch playlist btw. if you mean we should have it in BOTH switch_playlist and on program load then I"m perfectly happy with that

rexendevar:
> The playlist import code really needs to be its own function, something like `pctl.check_update_playlist_from_file()` . Because we will want to run this elsewhere, maybe even on startup, not just switch playlist.

done

rexendevar:
new additions:
- switched to pathlib for everything important. didn't do everything because it seemed like a lot more trouble than it was worth but I'm 90% sure it should be fully OS compatible
- auto-reloaded playlists now import their radio stations
- invalid playlist folder path now has a visual indicator (renders in a different color) and resets itself if the user closes the menu without fixing it
- attempted error handling for relative paths. can't test if it works but it should
- relative paths explainer is localized to user's OS and library
- deprecated "show" parameter in playlist export functions as it was never used
- if user drags in a playlist file that's already in the library, tauon will now switch to it

src/tauon/t_modules/t_prefs.py
```diff
@@ -322,6 +322,9 @@ class Prefs:
 	sync_playlist: int | None = None
 	download_playlist: int | None = None
 
+	autoscan_playlist_folder = False
```
C0rn3j: While it may look obvious, it's best to define the type for these two new vars too (bool and str)

src/tauon/t_modules/t_main.py
```diff
@@ -22207,36 +22388,47 @@ def __init__(self, tauon: Tauon) -> None:
...
 			if key not in ids:
 				del self.prefs.playlist_exports[key]
 
 	def render(self) -> None:
+		"""runs every frame that the playlist export menu is open.
+		also deals with the export entry logic."""
 		gui = self.gui
 		ddt = self.ddt
 		colours = self.colours
+		slash = "\\" if (self.tauon.system == "Windows" or self.tauon.msys) else "/"
```
C0rn3j, line 22423: Is there a reason to support playlist files with no extension?

Because if we didn't, we can throw out the entire logic with leaving slashes in string paths and further processing them, and we won't have to rely on string paths.

We could simply say that anything without an extension is a directory.

rexendevar: I was originally planning for anything without an extension to be a directory but Taiko said directory should be only when it ends in a slash and if it doesn't end in either then it's a file

rexendevar: also it's not supporting playlist files with no extension, this part is literally just handling the text displaying in the path box. every playlist has a relevant extension when it's exported, but we can't generate it in the body of the render function because then it appears every frame and it's a pain in the ass to edit the text

src/tauon/t_modules/t_prefs.py
rexendevar: will do 👍

rexendevar:
i wanna write the changelogs for this if that's alright btw

this will close #1612

rexendevar:
i suspect a couple edge cases with relative paths and network tracks interacting but it's not something i can test to see how or if it actually fails btw

Issue #1656, opened 2025-06-13
Trackpad: scrolling in certain views is extremely inconsistent


rexendevar:
**Desktop (please complete the following information):**
Fedora 42, Tauon development, from github actions
Gnome 48.2

**Describe the bug**
When viewing the unsynced lyrics showcase panel, the queue panel or the tracklist panel, scrolling vertically using a trackpad on modern GNOME produces unwanted behavior. On these panels ONLY (as far as I've seen):
1. The scroll speed appears to get some kind of acceleration?
2. The scroll distance is heavily quantized to a whole number of lines.
This means that when the physical scroll speed is slow, it's possible to scroll across the entire trackpad and not move the view at all.

Other views (synced lyrics when paused, unsynced mini lyrics, playlist panel) scroll smoothly with no visible processing. My other computer (Gnome 42.5)  does not get smooth scrolling in ANY view in Tauon and does not experience the same issue.

**Expected behavior**
Honestly if it's the same type of scrolling code just make everything scroll smoothly. get rid of all the scroll snapping, it's weird

**Screenshots/Videos**
https://youtu.be/EySJHPFrDLA

PR #1692, opened 2025-06-27
Fix smooth (non-integer) scrolling in most views

rexendevar:
A fix for [1656](https://github.com/Taiko2k/Tauon/issues/1656)


rexendevar:
if we need a function my thoughts are:
- inputs are scroll value, source and coefficient
- source is a string and allows the function to keep a dictionary of timers 
- maybe keep the function in its own class

rexendevar:
rebased it and made a new class for smooth scroll

Taiko2k:
Rebasing shouldn't add other commits, it makes it very hard to review.

rexendevar:
should i be rebasing against the current master and then squashing stuff? or against the initial state of the PR branch?

Taiko2k:
I feel we should do some renaming here for clarity. The old scroll variable can be renamed to `line_scroll` (`int`) (so yeah with an accumulator that increments `line_scroll` when it reaches 1/-1, and timeout). Parts that scroll by line can continue to use this.

Then we can have the class, maybe we should call it PixelScroll, (that outputs how many pixels to move for a given speed multiplier), because the term Smooth Scrolling traditionally refers to scrolling with acceleration/momentum (which we dont immediately want, though, in the future it would be nice to have natural scrolling via touch input (its an often requested feature), tauon has no touch support at current but it might be something to keep in mind if you're feeling big brained)

rexendevar:
I'm fine with renaming the class or something, but I think splitting it into multiple variables or functions is a bad and unnecessary move. The purpose of this whole thing is to combine standard integer scrolling (what occurs on GNOME 42.5 and idk what others) with smooth/float scrolling like what happens on newer versions, without adding any complexity in the implementation. Versions of Tauon prior to this change completely freak out when using float scrolling to go line by line, and the search overlay in particular is actually IMPOSSIBLE to scroll through in all practical terms; so simply "keeping the old implementation" for certain types of scrolling is not an option the way I see it.

I think replacing `self.inp.mouse_wheel` with `self.scroll.scroll` plus a random string that doesn't matter is as close to a drop-in replacement as you'll ever get. I'll admit some places may not need the new implementation, but it's actually the places with *pixel* scrolling, not line-by-line; and anyway I don't see a strong argument for removing it.

rexendevar:
Basically this should be used anywhere that scroll values must be integers, and I think I dropped it in everywhere I saw `int(self.inp.mouse_wheel)` without testing whether it actually needed to be an integer there

Taiko2k:
Ah I see the new class can function for line scrolling as well, thats fine

rexendevar:
it's ready to merge btw

C0rn3j:
> should i be rebasing against the current master

Yes.

PR #1693, opened 2025-06-28
Fix playlists with relative filepaths importing multiple copies of the same track

### LINKED ISSUE #1633: Various UI bugs and odd behaviors
 - OS: [e.g. Arch Linux/Windows/macOS] Arch
 - OS version: [e.g. rolling/11/15.2] rolling
 - Tauon Version: [e.g. 7.9.0, see Menu->Settings->About] 8.0.1
 - Installation method: [e.g. Flatpak/OS packager/portable binary/from source] AUR

**Describe the bug**
This is several. As I dug into my original bugs ( https://github.com/Taiko2k/Tauon/issues/1627 https://github.com/Taiko2k/Tauon/issues/1631 https://github.com/Taiko2k/Tauon/issues/1629 ) I found lots of behavior issues and bugs.

I noticed some of my pinned playlists had/have orange or yellow/green underlines every once in a while. What does that mean? I can't find anything in the docs about what these "underlines" might mean. I just had this happen again while testing some find in playlist functions but as soon as I try to screenshot it they go away. I finally caught it, what the hell does this mean?! What does it mean?! WHAT DOES IT MEAN!? - Jack Skellington
Image

Image

If you type for global search and ctrl click to add to the current playlist there is no feedback that it worked.

Using global search if you add a song with the "add artist" ctrl click and then add the same song "as the song" ctrl click you add the exact same file. One ignores tags and shows the filename and the other shows tags and has the play count...how can the same file have NULL plays and 50 plays? Why are they under different headers when the header values are the same? I suspect that "Song has 50 and NULL plays" bit might touch on my bug report about playlist queries being broken. https://github.com/Taiko2k/Tauon/issues/1629
Image

Here is another example of "Sense, this don't make it." Playlist sorted/grouped and I ran remove duplicates yet...there are clearly duplicates and somehow "the same files" are different...
Image

Find in playlist does not adapt to window sizes correctly. Tauon sits as a "little slit" on one of my screens. 360px tall by 1280 wide. If I do ctrl F and type it often buries the "jumped to" and highlighted match below what you can see.

ctrl F in a playlist also does not always match. I just did some test searches and while I know it was spelled correctly it wouldn't allow me to hit enter despite not being red. This then triggered an inconsistency in search clearing behavior. I would have to backspace/delete everything before I could do a new search. Even if I hit esc and ctrl F again it would come up with the previous search terms still there. I had this happen several times but I can't seem to trigger it consistently to give more concise explanation of when/why the behavior happens.

I just did a test search for one of my songs "Olde Tyme tError" and when I hit enter it played the correct song but in the playlist view it highlighted Pop-Stars by KDA. At times I will type something correctly and hit enter and it just does nothing. You can see in the below screenshot two songs are highlighted as selected, one highlighted as playing.
Image

You may also notice there are some songs that look "grayed out." Those where added to the playlist before a folder rename. So they are "missing." However I added them to the playlist with the new directory. Now if I go to where the new ones are and play them, they play. However if I ctrl F and search for one of those songs names, despite the playlist having -valid- paths to a matching file it finds the dead ones, immediately exits the search before it finds the valid matches and just plays the "next song." So if you have missing songs in your playlist and don't notice your searches will be adversely affected.

Man ctrl F is getting the screws today! Ctrl F type something, spelling is perhaps wrong or there is no match in the current playlist...it goes red. OK great it's red for bad but if I hit enter it does nothing. I have to hit esc and then ctrl f again to clear it or backspace/delete everything typed...hitting enter should just clear it after you've already indicated no match so the user can quickly retry the spelling or look for something else.

**To Reproduce**
1. Install Tauon
2. GoTo 10

**Expected behavior**

**Screenshots/Videos**

**Tauon log**

**Additional context**
I found other issues but I don't want to post -yet-. There is a degree of worried I'm being PEBKAC.

I want to note that hopefully my bug reporting style (and volume) doesn't make you guys rip your hair out. I've been looking for a music player with specific features for at least 20 years to rival foobar2000. I've tried every damn *nix player there was since the mid 1990's and Tauon is the first one that might take the crown...at least for me. So Thank You! I hope this flood/influx of issues isn't taken as pooping on what stands to be a first class player.

rexendevar:
partially fixes #1633. I'm also planning to update the remove duplicates function to fix past imports

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -6519,7 +6519,6 @@ class Tauon:
 						line = Path(pl_dir / Path(line) ).resolve()
 					else:
 						line = Path(line).resolve()
-					logging.info(f"final code says {line}")
 					line = str(line)
 
 					# Cache datbase file paths for quick lookup
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -6487,6 +6487,7 @@ class Tauon:
 
 		if not os.path.isfile(path):
 			return
+		pl_dir = Path(path).parent
 
 		with Path(path).open(encoding="utf-8") as file:
 			lines = file.readlines()
@@ -6513,17 +6514,13 @@ class Tauon:
 						self.radiobox.start(radio)
 				else:
 					line = uri_parse(line)
-					# Join file path if possibly relative
-					if not line.startswith("/"):
-						line = os.path.join(os.path.dirname(path), line)
-
-					# Get rid of redundant relative paths
-					split = line.split("/")
-					while ".." in split:
-						location = split.index("..")
-						del split[location]
-						del split[location-1]
-					line = "/".join(split)
+					# Fix up relative filepaths
+					if not Path(line).is_absolute():
+						line = Path(pl_dir / Path(line) ).resolve()
+					else:
+						line = Path(line).resolve()
+					logging.info(f"final code says {line}")
+					line = str(line)
 
 					# Cache datbase file paths for quick lookup
 					if not location_dict:
@@ -6631,6 +6628,7 @@ class Tauon:
 			a: list[dict[str, str | None]] = []
 			b: dict[str, str | None] = {}
 			info = ""
+			pl_dir = Path(path).parent
 
 			for top in e:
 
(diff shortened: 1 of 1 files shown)


rexendevar:
need to replace the whole thing with a pathlib implementation

Issue #1699, opened 2025-06-29
prevent duplicates of the same column in columns mode


rexendevar:
**Is your feature request related to a problem? Please describe.**
I use a bunch of columns all the time. I noticed just the other day that I actually have three copies of the filepath column and two each of genre, release date, duration, plus a rating column that i don't use and a playcount when i prefer scrobble count. there was a shit ton of clutter out of view.

**Describe the solution you'd like**
Instead of having plus signs by every column in the list, we should have checkmarks for ones that are enabled.

Issue #1696, opened 2025-06-29
Tauon track ID option in columns mode


rexendevar:
Would be useful to have the track ID next to the track when I'm debugging. probably could add this myself but not right now lol

C0rn3j:
I'd probably at these to Track Info menu and only on when SHIFT is held.

Or you can show it when debug flag is on (which is currently a file within the userdata folder called `debug`, or lack of it for off state)

rexendevar:
What about the option only appears in the list of possible columns if you hold shift while it's open?

PR #1707, opened 2025-07-03
Some QOL changes mainly for columns mode

rexendevar:
- added "internal track ID" as an optional column under More
- columns are now selectable as a checkbox. removed ability to have duplicate columns
- widened default date column to fit full date format (not just YYYY)
- new icons for lyrics in columns mode - checkmark for lyrics and hourglass for synced lyrics
- better detection in find_synced_lyrics_data 
- find_synced_lyrics_data now writes to track.synced and runs during rescan_tags


rexendevar:
completes #1696 and #1699

PR #1729, opened 2025-07-14
option to manually edit lyrics

### LINKED ISSUE #1723: [Feature Request] Write lyrics into the file's metadata
I use the same music files between my pc and phone, when I add lyrics through Tauon they are not synced between devices as they are not added into the file's metadata, I end up having to edit the lyrics again on another program so that I can have the lyrics show up on my phone too.

ImageImage
Image

Ideally when I search or paste lyrics, I'd like them to be automatically written into the metadata.

**Describe alternatives you've considered**
Alternativelly there could be a menu option to do it manually.

rexendevar:
- new "edit lyrics" button option under "misc" when right clicking any lyrics
- new option in settings to save all lyrics changes back to original file (resolves #1723)

concerns/possible TODOs:
- add cancel [X] button beside "reload" ?
- how to deal with synced lyrics being returned from `get_lyric_fire`? do we write them to the normal `USLT` lyrics tag or do we figure out how to support `SYLT`?
- ~~is bare ID3 enough?? I struggled with Mutagen~~ FIXED for flac and m4a - ogg and ape are untested
- future big project - allow manual syncing of lyrics

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -564,6 +564,7 @@ class GuiVar:
 
 		self.combo_mode = False
 		self.showcase_mode = False
+		self.timed_lyrics_edit_view: bool = False
 		self.display_time_mode = 0
 
 		self.pl_text_real_height = 12
@@ -5746,6 +5747,7 @@ class Tauon:
 		self.tab_menu              = Menu(self, 160, show_icons=True)
 		self.playlist_menu         = Menu(self, 130)
 		self.showcase_menu         = Menu(self, 135)
+		self.timed_lyrics_edit_menu= Menu(self, 135)
 		self.spotify_playlist_menu = Menu(self, 175)
 		self.queue_menu            = Menu(self, 150)
 		self.radio_entry_menu      = Menu(self, 125)
@@ -13058,10 +13060,11 @@ class Tauon:
 			self.gui.combo_mode = False
 			self.gui.was_radio = False
 
-	def enter_showcase_view(self, track_id: int | None = None) -> None:
+	def enter_showcase_view(self, track_id: int | None = None, timed_lyrics_edit: bool=False) -> None:
 		if not self.gui.combo_mode:
 			self.enter_combo()
 			self.gui.was_radio = False
+		self.gui.timed_lyrics_edit_view = timed_lyrics_edit
 		self.gui.showcase_mode = True
 		self.gui.radio_view = False
 		if track_id is None or self.pctl.playing_object() is None or self.pctl.playing_object().index == track_id:
@@ -13078,6 +13081,9 @@ class Tauon:
 		self.gui.radio_view = True
 		self.inp.mouse_click = False
 		self.gui.update_layout = True
+
+	def enter_timed_lyrics_edit(self, track: TrackClass) -> None:
+		self.enter_showcase_view(track.index, timed_lyrics_edit=True)
 
 	def standard_size(self) -> None:
 		self.prefs.album_mode = False
@@ -35546,10 +35552,14 @@ class Showcase:
 		self.guitar_chords = tauon.guitar_chords
 		self.showcase_menu = tauon.showcase_menu
 		self.scroll        = tauon.smooth_scroll
+		self.timed_lyrics_edit = TimedLyricsEdit(tauon=tauon)
 		self.lastfm_artist = None
 		self.artist_mode = False
 
 	def render(self) -> None:
+		if self.gui.timed_lyrics_edit_view:
+			self.timed_lyrics_edit.render()
+			return
 		box = int(self.window_size[1] * 0.4 + 120 * self.gui.scale)
 		box = min(self.window_size[0] // 2, box)
 
@@ -36719,6 +36729,8 @@ class SmoothScroll:
 
 			return scroll_distance
 
+
+
 class TimedLyricsEdit:
 	def __init__(self, tauon: Tauon) -> None:
 		self.tauon         = tauon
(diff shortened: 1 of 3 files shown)


Taiko2k:
~~I don't think we need any extra buttons for this. Perhaps a message box is sufficient. e.g. something like "*File opened. Save and close editor to save changes. Lyrics will be saved to database / written to tags* " (depending on setting).~~

~~Probably don't need any variables to track this at all. This could all be done in one blocking thread. So, once the thread is finished, it cleans itself up. You could also edit multiple at a time.~~

Taiko2k:
Actually afaik we don't have a way to directly track if the user closed the editor. Hmm.

rexendevar:
yeah i copied the code from the bit that deals with manually editing the config file. i figured if you couldn't figure out a simpler way to save it then i probably couldn't either

src/tauon/t_modules/t_main.py
```diff
@@ -16837,6 +16925,15 @@ def save_state(self) -> None:
 			with (self.user_directory / "lyrics_substitutions.json").open("w") as file:
 				json.dump(prefs.lyrics_subs, file)
 
+			edited_lyrics = Path( self.config_directory / _("edited-lyrics") )
```
C0rn3j: This is already a Path object, no need to wrap it again

src/tauon/t_modules/t_main.py
```diff
@@ -35780,6 +35915,12 @@ def render(self) -> None:
 					w,
 					int(self.window_size[1] - 100 * self.gui.scale),
 					0)
+
+				lyric_file = Path( self.tauon.config_directory / _("edited-lyrics") / str( self.pctl.track_queue[self.pctl.queue_step] )).with_suffix(".txt")
+				if self.gui.opened_lyric_file and lyric_file.is_file():
+					if Drawing(tauon=self.tauon, pctl=self.pctl).button( _("Load lyrics"), x - 50*self.gui.scale - self.ddt.get_text_w(_("Load lyrics"), 211), int(self.window_size[1] - 100 * self.gui.scale), tooltip=_("Make sure to save your changes."), background_colour = ColourRGBA(90, 50, 130, 255)):
```
Taiko2k: Instantiating the Drawing class each use isn't the intended way of doing it. We use the ~~`self.ddt`~~ `tauon.draw` instance.

rexendevar: I don't remember where exactly this is, but there were some places where self.ddt never got instantiated or maybe couldn't be. i don't remember. other places use self.ddt

rexendevar: oh wait no I'm using self.tauon.draw for all of them. i don't remember why but it doesn't add any tech debt far as I'm concerned

Taiko2k: ah right its `draw`, so use `tauon.draw()`  instead of `Drawing()`

rexendevar: i swear there must have been a reason why i didn't do that to begin with. i guess I'll check whether i was wrong later lol

rexendevar:
Still need to add support for ogg, opus, oga, wv, tta, ape and mp4 formats!!!!!! I don't know how to write those

src/tauon/t_modules/t_main.py
```diff
@@ -7642,6 +7675,84 @@ def clear_lyrics_deco(self, track_object: TrackClass) -> list[ColourRGBA | None]
...
 		return [line_colour, self.colours.menu_background, None]
 
+	def edit_lyrics_deco(self, track_object: TrackClass) -> list[ColourRGBA | None]:
+		line_colour = self.colours.menu_text# if track_object.lyrics else self.colours.menu_text_disabled
+
+		return [line_colour, self.colours.menu_background, None]
+
+
+	def edit_lyrics(self, track_object: TrackClass) -> None:
+		target = Path( self.config_directory / _("edited-lyrics") / str(track_object.index)).with_suffix(".txt")
```
Taiko2k, line 7685: I don't think config directory is quite right here. Maybe change `self.config_directory` to `self.temp_directory`... ok that doesn't exist yet but `tmp_cache_dir()` does. Not sure why I made that a function like that.

src/tauon/t_modules/t_main.py
```diff
@@ -7642,6 +7675,84 @@ def clear_lyrics_deco(self, track_object: TrackClass) -> list[ColourRGBA | None]
...
+		if not target.parent.is_dir():
+			target.parent.mkdir()
+			# TODO (Flynn): delete all lyric files on program close
+		with open(target, "w") as lyrics_file:
+			if track_object.lyrics:
+				# TODO (Flynn): make this interchangablaedbaledjbaejdblaejdblajebjd synced lyrics too lol
+				if find_synced_lyric_data(track_object, just_check=True):
+					lyrics_file.write( _("NOTE: You are editing synced lyrics. Please don't break the formatting, because we can't correct for that yet. Thanks") + '\n')
```
Taiko2k, line 7693: `track_object.lyrics` should never be synced. If we are to edit synced lyrics this way I think this needs to be a clearly distinct button/function.

Also im not sure about writing messages like this because the user would need to remove them before saving.

My advice would be to remove synced lyrics from this function completely.

src/tauon/t_modules/t_main.py
```diff
@@ -7642,6 +7675,84 @@ def clear_lyrics_deco(self, track_object: TrackClass) -> list[ColourRGBA | None]
...
+			audio["Lyrics"] = track.lyrics
+			audio.save()
+			logging.info(f"Edited lyrics in the file for {track.artist} - {track.title}")
+		elif track.file_ext in ("MP4","M4A","M4B","M4P"):
+			audio = mutagen.mp4.MP4(track.fullpath)
+			audio['\xa9lyr'] = track.lyrics
+			audio.save()
+			logging.info(f"Edited lyrics in the file for {track.artist} - {track.title}")
+		else:
+			logging.info(f"Could not save lyrics to file with extension {track.file_ext}")
```
Taiko2k, line 7734: We need a GUI facing error message for this failure. I don't think we should fail silently.

src/tauon/t_modules/t_main.py
```diff
@@ -41271,9 +41438,10 @@ def get_xcursor(name: str):
 
 	center_info_menu.add_sub(_("Misc…"), 150)
 	center_info_menu.add_to_sub(0, MenuItem(_("Substitute Search..."), tauon.show_sub_search, pass_ref=True))
-	center_info_menu.add_to_sub(0, MenuItem(_("Paste Lyrics"), paste_lyrics, tauon.paste_lyrics_deco, pass_ref=True))
-	center_info_menu.add_to_sub(0, MenuItem(_("Copy Lyrics"), copy_lyrics, tauon.copy_lyrics_deco, pass_ref=True, pass_ref_deco=True))
-	center_info_menu.add_to_sub(0, MenuItem(_("Clear Lyrics"), clear_lyrics, tauon.clear_lyrics_deco, pass_ref=True, pass_ref_deco=True))
+	center_info_menu.add_to_sub(0, MenuItem(_("Paste Lyrics"), tauon.paste_lyrics, tauon.paste_lyrics_deco, pass_ref=True))
```
Taiko2k: What if the user is/isn't in the showcase view, do they see the "load lyrics" button? Because there's 2 places the lyrics options can appear, the side panel and the showcase, but only one button? It might be better to do a `show_message()` "Read lyrics back from the file?"  with a callback func instead of a button.

Taiko2k (changes requested):

src/tauon/t_modules/t_main.py
rexendevar: That might work. I did make sure that the leftover files get deleted on program close, if that's what you're worrie about

src/tauon/t_modules/t_main.py
rexendevar: On your second point, the reload function strips out that exact text. And `track_object.lyrics` easily can be synced, if the metadata field was written by a different program

src/tauon/t_modules/t_main.py
rexendevar: The button appears regardless of which view you're in

rexendevar: but if I'm honest, I think I want to do a couple more days' work on the synced lyrics editor and then ditch the text file system entirely. i only really need the backend from this pr

src/tauon/t_modules/t_main.py
rexendevar: fair

src/tauon/t_modules/t_main.py
rexendevar: actually no i think i'm wrong about this. synced and unsynced lyrics get read from the same metadata field right

src/tauon/t_modules/t_main.py
```diff
@@ -19751,8 +19881,20 @@ def render(self, index: int, x: int, y: int, side_panel: bool = False, w: int =
 				association = collider, line
 				line_positions.append( association )
 
+		button_pressed = False
+		lyric_file = Path( self.tauon.config_directory / _("edited-lyrics") / str( self.pctl.track_queue[self.pctl.queue_step] )).with_suffix(".txt")
```
C0rn3j: config_directory is Path already, if you need to wrap with brackets to do a function call on it, do `()` instead of `Path()`.

This is present in other parts of the code too.

rexendevar:
decided to just incorporate the changes into #1762

PR #1762, opened 2025-08-03
Synced lyrics editor

### LINKED ISSUE #1690: Need To Stop And Restart To Get Lyrics Playing
Hi,

I'm running CachyOS, with Hyprland, and Tauon, 8.0.1

I noticed when I start to play a song, and then search for lyrics, once the lyrics are done downloading, it won't play them.

I have to then close and reopen Tauon, and then I'll see the lyrics being highlighted, and moving to the next line of lyrics being sung, and the lyrics plays along for you.

This never works when first downloading the lyrics, I always have to restart to make it work.

THANKS

### LINKED ISSUE #1723: [Feature Request] Write lyrics into the file's metadata

rexendevar:
Adds a full-featured synced lyrics editor. 

Marking the pull as complete, under the condition that the lrclib-solver binary will be located at `/app/bin/src/tauon/lrclib-solver` in the Flatpak release once Taiko gets around to it.

Changes made outside of the new editor view include:
- `pctl.jump_time` now works with millisecond precision and has been typed as a float
- made an attempt to fix button text offset given different font sizes. it doesn't work.
- `sync_lyrics_time_offset` now defaults to -150, and has been implemented in timedlyricsren
- all lyrics changes now propagate back to their files (resolves #1723)
- blank lines now display as music notes in timedlyricsren
- optional argument for `pctl.play()` and `pctl.stop()` prevents them from refreshing the gui
- optional arguments for `find_synced_lyric_data` do things that probably aren't very useful to anyone but me
- rescanning tags now also reloads lyrics
- new gui variable `now_searching` tracks the status of lyrics searches
- new input variable checks if the S key is pressed
- lyrics searches from LRCLIB are now passed the correct user agent
- made an attempt to resolve #1690
- reworded toggle box for the playlist folder auto rescan option in settings. completely unrelated but i couldn't be bothered making a new pr.
- typing/exception fixes in a couple places

Still left to do:
 - [x] remove a ton of tasks i did ages ago
 - [x] make upload menu look better
 - [x] fix scroll timestamp scrolling 
 - [x] figure out call path for lrclib challenge solver
 - [x] compatibility with martin's backend work
 - [x] compatibility with network tracks
 - [x] cleanup
 - [x] write a tutorial
 - [ ] add us to wikipedia
 - [x] (on a new commit) support uploading to lrclib


src/tauon/t_modules/t_main.py
```diff
@@ -5954,6 +5956,8 @@ def __init__(self, holder: Holder, bag: Bag, gui: GuiVar) -> None:
 		self.playlist_autoscan = False
 		self.dropped_playlist = -1
 
+		self.now_searching: int = 0
```
C0rn3j: I see this goes from -1 to 2 or such, what does it actually do in those states?

src/tauon/t_modules/t_main.py
```diff
@@ -13058,10 +13066,11 @@ def exit_combo(self, restore: bool = False) -> None:
 			self.gui.combo_mode = False
 			self.gui.was_radio = False
 
-	def enter_showcase_view(self, track_id: int | None = None) -> None:
+	def enter_showcase_view(self, track_id: int | None = None, timed_lyrics_edit: bool=False) -> None:
```
C0rn3j: Nitpick: `bool=False` -> `bool = False`

src/tauon/t_modules/t_main.py
```diff
@@ -36720,6 +36738,784 @@ def scroll(self, source: str, coeff: float | int = 1) -> int:
...
+		self.renderer      = tauon.renderer
+		self.lyrics_ren    = tauon.lyrics_ren
+		self.window_size   = tauon.window_size
+		self.scroll        = tauon.smooth_scroll
+		self.fields        = tauon.fields
+		self.lastfm_artist = None
+		self.artist_mode = False
+		self.recenter_timeout = Timer()
+		self.overlay       = tauon.search_over
+
```
C0rn3j: Nitpick; move .fields and .search_over up there with the rest sorted by length

src/tauon/t_modules/t_main.py
```diff
@@ -36720,6 +36738,784 @@ def scroll(self, source: str, coeff: float | int = 1) -> int:
...
+		self.struct_track: int = -1 #what track are we on
+		self.structure: list[ tuple[ str, float, str] ] = []
+		# each line contains a timestamp as a string, that timestamp's actual time, and the line itself
+		self.line_edit_box = TextBox2(tauon=tauon)
+
+		self.temp_w: int = 0
+		self.temp_scale: float = self.gui.scale
+		self.temp_line: int = -1
+
+		self.dont_scroll: bool = False #cancels auto scroll correction when adding & removing lines
```
C0rn3j: It'd probably be best to avoid the double negative the naming of the var gives

src/tauon/t_modules/t_main.py
```diff
@@ -36720,6 +36738,784 @@ def scroll(self, source: str, coeff: float | int = 1) -> int:
...
+		self.font = 20
```
C0rn3j: Missing type

src/tauon/t_modules/t_main.py
```diff
@@ -36720,6 +36738,784 @@ def scroll(self, source: str, coeff: float | int = 1) -> int:
...
+		self.menu.add(MenuItem(_("Search for Lyrics"), tauon.get_lyric_wiki, tauon.search_lyrics_deco, pass_ref=True, pass_ref_deco=True))
+		self.menu.add(MenuItem(_("Clear All Timestamps"), self.clear_all_timestamps, pass_ref=False))
+		self.menu.add(MenuItem(_("Clear All Lyrics"), self.clear_lyrics, pass_ref=False))
+
+
+
+	def button(self, text: str, x_pos: int, y_pos: int, font: int,
+			bg: ColourRGBA | None = None, active_bg: ColourRGBA | None = None,
+			txt: ColourRGBA | None = None, active_txt: ColourRGBA | None = None,
+			tooltip: str = "", off: bool = False) -> bool | None:
```
C0rn3j: Set your tab-width to a different size to see the issue with this one, you're using indentation for visual alignment here, which will fail.

You can put the args starting with `self` immediately on a new line

src/tauon/t_modules/t_main.py
```diff
@@ -36720,6 +36738,784 @@ def scroll(self, source: str, coeff: float | int = 1) -> int:
...
+			s += int(ms) / 100
+		elif len(ms) == 3:
+			s += int(ms) / 1000
+		return s
+
+
+
+	def get_stamp_from_time(self, t: float) -> str:
+		if t<0:
+			return _("?????")
```
C0rn3j: No need to translate question marks... right¿

Or is this a placeholder?

src/tauon/t_modules/t_main.py
```diff
@@ -36720,6 +36738,784 @@ def scroll(self, source: str, coeff: float | int = 1) -> int:
...
+			self.scroll_position -= self.yy
+			self.dont_scroll = True
+		self.gui.pl_update += 1
+
+
+
+	def save(self) -> None:
+		lyrics = ""
+		warning = [False, False]
+		timed = 0
```
C0rn3j: Missing type

src/tauon/t_modules/t_main.py
```diff
@@ -36720,6 +36738,784 @@ def scroll(self, source: str, coeff: float | int = 1) -> int:
...
+			self.scroll_position = scroll_to
+			#drop_w, self.line_height = self.ddt.get_text_wh("?", font_size, 100, True) # get height of one line
+			self.temp_scale = self.gui.scale
+			self.temp_w = w
+			self.line_height = self.ddt.get_text_w("?", self.font, True)
+
+		self.yy = self.line_height + spacing
+		# don't autoscroll if the new active line is not visible
+		if ( self.scroll_position > self.window_size[1]/2 or self.scroll_position < -self.window_size[1]/2 or self.editing_text ) \
+			and self.temp_line != line_active and not self.dont_scroll:
```
C0rn3j: I'd personally put the `and` on the same indent-width as the if

src/tauon/t_modules/t_main.py
rexendevar: 1 means currently running, -1 means exited with error, 2 means success (and we should reload the lyrics)

src/tauon/t_modules/t_main.py
rexendevar: i have to delete a few of these actually cause I'm not using them anymore but will do

src/tauon/t_modules/t_main.py
rexendevar: i figure it's better safe than sorry, especially with stuff like Chinese maybe. worst case all the translations just say ?????

src/tauon/t_modules/t_main.py
C0rn3j: That should probably be an enum then, for readability, you can check out my latest PR since I converted parts of the codebase to them

rexendevar: instead of an enum can i just declare it as one of 4 possible strings? something like `self.now_searching: "off" | "running" | "errored" | "success" = "off"` ? doing a full enum for this one feature seems like overkill but idk how typing works

C0rn3j: I think you could do a Literal in that manner

https://typing.python.org/en/latest/spec/literal.html

Though whether it's a good idea is questionable.

Either way is going to be better than just raw magic numbers, though.

rexendevar: I think as long as it's declared so it can be used later  it should be fine

src/tauon/t_modules/t_main.py
```diff
@@ -36720,6 +36738,894 @@ def scroll(self, source: str, coeff: float | int = 1) -> int:
...
+		self.allow_scroll: bool = True #cancels auto scroll correction when adding & removing lines
+		self.font: int = 20
+		self.line_height = round(self.ddt.get_text_w("?", self.font, True))
+		self.yy: int = self.line_height + round(10 * self.gui.scale) #line height plus spacing
+
+		self.autosave_timer = Timer()
+		self.autosaved: bool = True
+		self.cursor: int | None = None
+		self.text_timer = Timer()
+		self.x_posns: list[int]=[] # to display buttons etc in the right places
```
C0rn3j: Nitpick: `]=[]` -> `] = []`

src/tauon/t_modules/t_main.py
```diff
@@ -7639,6 +7682,69 @@ def clear_lyrics_deco(self, track_object: TrackClass) -> list[ColourRGBA | None]
...
+				audio['\xa9lyr'] = lyrics
+				audio.save()
+				logging.info(f"Edited lyrics in the file for {track.artist} - {track.title}")
+			else:
+				self.show_message(
+					_("Could not write lyrics to file"),
+					_("We don't know how to write lyrics to the filetype: ") + track.file_ext,
+					mode="error"
+				)
+		except:
```
C0rn3j, line 7738: Bare excepts shouldn't exist, `except Exception` it, and add a `logging.exception` in it, so the exception gets logged

src/tauon/t_modules/t_prefs.py
```diff
@@ -187,7 +188,7 @@ class Prefs:
 
 	guitar_chords = False
 	prefer_synced_lyrics = True
-	sync_lyrics_time_offset = 0
+	sync_lyrics_time_offset = -500
```
C0rn3j: Typing please

src/tauon/t_modules/t_main.py
```diff
@@ -19378,11 +19499,13 @@ def __init__(self, tauon: Tauon, pctl: PlayerCtl) -> None:
 	def button(
 		self, text: str, x: int, y: int, w: int | None = None, h: int | None = None, font: int = 212, text_highlight_colour: ColourRGBA | None = None, text_colour: ColourRGBA | None = None,
 		background_colour: ColourRGBA | None = None, background_highlight_colour: ColourRGBA | None =None, press: bool | None = None, tooltip: str="") -> bool:
+		'''PSA for anyone making a new button function: use fields.add(rect) to make the gui
```
C0rn3j, line 19502: We use double quotes for docstrings - this issue is in 6 more places from what I see

C0rn3j:
Path suggestion:

* On Linux `{installdir}/lrclib-solver`, if that is not found, try `/bin/lrclib-solver` too
* On macOS `{installdir}/lrclib-solver`
* On Windows `{installdir}/lrclib-solver.exe`

src/tauon/t_modules/t_lyrics.py
```diff
@@ -131,6 +131,18 @@ def lrclib(artist: str, title: str) -> tuple[str, str]:
 			return p, s
 	return "", ""
 
+def get_lrclib_challenge() -> tuple[str]:
+	h = {
+		"User-Agent": "TauonMusicBox/Devel",
```
C0rn3j: This should use the Tauon version now

src/tauon/t_modules/t_main.py
```diff
@@ -36768,6 +36960,1536 @@ def scroll(self, source: str, coeff: float = 1) -> int:
...
+		if not target.is_dir():
+			return
+		for child in target.iterdir():
+			if child.is_file():
+				count += 1
+				child.unlink()
+		try:
+			target.rmdir()
+		except:
+			logging.error( _("You put a folder in the lyrics-editor directory. Don't do that.") )
```
C0rn3j: except should be specific + have a generic case for unknown exceptions.

It looks like it also makes no sense to have that here, as the check above already makes sure this will never be a directory due to the return?

src/tauon/t_modules/t_main.py
```diff
@@ -36768,6 +36960,1536 @@ def scroll(self, source: str, coeff: float = 1) -> int:
...
+		self.reload_menu()
+
+
+
+
+	def button(
+			self, text: str, x_pos: int, y_pos: int, font: int,
+			bg: ColourRGBA | None = None, active_bg: ColourRGBA | None = None,
+			txt: ColourRGBA | None = None, active_txt: ColourRGBA | None = None,
+			tooltip: str = "", off: bool = False, return_rect: bool = False) -> bool | None:
```
C0rn3j, line 37175: You're returning a tuple below too

src/tauon/t_modules/t_main.py
```diff
@@ -36768,6 +36960,1536 @@ def scroll(self, source: str, coeff: float = 1) -> int:
...
+				full_line = ( self.get_stamp_from_time(time), time, self.structure[self.line_active+1][2] ) # else time the next line
+				self.structure[self.line_active+1] = full_line
+			self.scroll_position -= self.yy
+			self.allow_scroll = False
+		self.queue_next_frame = True
+
+	def scroll_timestamp(self, current_line: int, active: bool = True) -> bool:
+		stamp, time, line = self.structure[current_line]
+		if time == -1.0:
+			return
```
C0rn3j: This returns None, not bool

src/tauon/t_modules/t_main.py
```diff
@@ -36768,6 +36960,1536 @@ def scroll(self, source: str, coeff: float = 1) -> int:
...
+		pasted_lines = self.line_edit_box.text.splitlines()
+		temp_line = self.structure[current_line]
+		overwrite: bool = (self.inp.key_lalt or self.inp.key_ralt)
+
+		def is_int(number: str) -> bool:
+			try:
+				int(number)
+				return True
+			except:
+				return False
```
C0rn3j: bare except

src/tauon/t_modules/t_main.py
rexendevar: is there any way i can return two separate things in some cases while still allowing the button be used in if statements in others?

C0rn3j: I'm not sure about the code flow here and am too tired to check, but what I meant is that the return typing is wrong

src/tauon/t_modules/t_main.py
```diff
@@ -36777,6 +36973,1591 @@ def scroll(self, source: str, coeff: float = 1) -> int:
...
+
+		# nudge timestamp
+		self.check_timer        = Timer() # preview timing after half a second
+		self.check_line:    int = -1
+		self.check: bool | None = None    # True if previewing now, False if waiting to preview, None if disengaged
+
+		# lrclib upload
+		self.upload_synced:               bool = True
+		self.upload_static:               bool = False
+		self.potential_uploads: dict[int:dict] = {}
```
C0rn3j, line 37041: Did you mean to do `dict[int, dict]` here?

It would still be good to type the dict itself.

I.e. `dict[int, dict[str, str | float]]` or whatever the actual typing is.

src/tauon/t_modules/t_main.py
```diff
@@ -39226,6 +41015,26 @@ def cache_paths() -> tuple[dict, dict]:
 				dic2[(value.fullpath.replace("\\", "/"), value.track_number)] = value
 		return dic, dic2
 
+	def upload_to_lrclib( upload: dict ) -> bool:
```
C0rn3j: It would be ideal to type the dict too

PR #1771, opened 2025-08-07
Send "Seeked" signal on track repeat

### LINKED ISSUE #1756: Send "Seeked" signal on track repeat/restart
When a track ends it doesn't send any kind of signal to DBUS and causes event-driven extra apps to not sync with the track position correctly.

Other players do this by sending a "Seeked" signal when the song ends and goes to beginning because of repeat mode

BEST8OY:
Fixes #1756

Code Generated Via Gemini Agent Mode

### DIFF
--- a/src/tauon/t_modules/t_dbus.py
+++ b/src/tauon/t_modules/t_dbus.py
@@ -445,6 +445,12 @@ class Gnome:
 									logging.debug(track.fullpath)
 
 					def update_progress(self) -> None:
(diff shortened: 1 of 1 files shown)


C0rn3j:
The two new variables should be added to the `__init__` function of the class along with their type.

i.e.

```python
self.last_playing_time: float = 0.0
self.last_track_index: int = -1
```

C0rn3j (approved):

Taiko2k:
`pctl.playing_object()` may return `None` so we need to check for that before continuing.

C0rn3j (commented):
Make sure you have indentation visible in your editor, this won't run

image

Issue #1316, opened 2024-11-25
Migrate to `org.kde.StatusNotifierItem` from `libayatana-appindicator`/`libappindicator-gtk3`


C0rn3j:
https://github.com/AyatanaIndicators/libayatana-appindicator/issues/22

Tauon cannot use GTK4 before this is done as the system tray feature relies on GTK3 to be available, and mixing 3.0 and 4.0 is not possible.

Example project that has done this already is Safe Eyes - https://github.com/slgobinath/SafeEyes/pull/558

stkw0:
How it would work for people not using KDE (nor GNOME)?

C0rn3j:
The DE needs to implement the spec - https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/ - even Xfce seems to do so based on a quick check, so I imagine everyone supports it one way or another.

stkw0:
I was thinking specifically with people who uses a window manager instead of a DE.

C0rn3j:
Window managers tend to have tray support?

Either way, same deal, instead of AppIndicator spec you need the StatusNotifierItem spec implemented.

stkw0:
Yes. At least dwm did, and most wm I know or used had it (though I moved to KDE years ago). In any case, I guess it could be made in such a way that notifications are optional and Tauon can run without notifications support if the WM don't implement that spec.

C0rn3j:
Note that this is not about notifications but the tray icon and its associated functions.

Tray icon is already optional and off by default.

Both the current indicator and the kde implementation over dbus(?) actually are both SNI (StatusNotifierItem) implementations... as you can see I have not looked too deep into this

C0rn3j:
Dependent on #1332, but https://www.phoronix.com/news/SDL-System-Tray-Support seems to perhaps be the best option here. 

We'd ditch all the current tray implementations for an SDL native one, since Tauon is heavily based on SDL anyway.

C0rn3j:
`(tauonmb:608088): libayatana-appindicator-WARNING **: 02:07:58.823: libayatana-appindicator is deprecated. Please use libayatana-appindicator-glib in newly written code.`

New warning too.

C0rn3j:
We have now migrated to SDL native functions for the tray, so this will be automatically fixed when SDL 3.6.0 releases.

Keeping this open to remember to bump GTK 3 to GTK 4 (or tear it out completely) when that happens.

PR #1772, opened 2025-08-07
Implement native SDL tray

C0rn3j:
Will eventually fix these issues - #1316 and #1904 and #2137, after SDL gets a better tray implementation - see https://github.com/libsdl-org/SDL/pull/14582 and https://github.com/libsdl-org/SDL/pull/15189 and https://github.com/libsdl-org/SDL/issues/15246 about that.

Some reading material is also in https://github.com/libsdl-org/SDL/issues/11703 and https://github.com/libsdl-org/SDL/issues/13666

Commits with the ayatana glib implementation were removed, in case that's actually desired later for some odd reason - [1772.patch](https://github.com/user-attachments/files/24033562/1772.patch)

~~Currently it seems to be blocked by https://github.com/AyatanaIndicators/libayatana-appindicator-glib/issues/87~~

~~We could still improve on this by using SDL instead.~~

- [ ] Get show/hide working - waiting on SDL response to see if this will be possible - meanwhile I implemented destroying of the tray upon show/hide.
- [ ] Get accessiblity strings working - waiting on SDL response to see if this will be possible - not too important
- [ ] Get scroll up/down to change volume - waiting on SDL implementation - not too important, we can ship without it for now
- [ ] Implement tooltip description - we can keep the app title and put the song info into the description if this gets implemented in SDL - https://github.com/libsdl-org/SDL/issues/15246#issuecomment-4467050199
- [ ] Tray title is broken on Plasma - https://github.com/libsdl-org/SDL/issues/15618
- [x] Get Windows/macOS icons sorted out - do we even have a tray on macOS? On Windows we'll need to tear out lynxtray
- [x] Implement `sdl_tray_set_label` / `sdl_tray_set_title` - *set_label removed and sdl_set_tray_tooltip now used for title*


eafton:
Sorry if this is taking so long, I am waiting for others to give feed back on https://github.com/libsdl-org/SDL/issues/15246 before I can implement these features you need.

C0rn3j:
@eafton No worries mate, I am glad that it's being resolved at all!

@Taiko2k I think you force pushed the wrong thing?

C0rn3j:
Can be pulled as-is (please squash).

Note that:
* the SNI rework for SDL is still not in the current SDL version (presumably will be out with the 3.6.0 release)
* setting tray tooltip is broken on Plasma (which seems to be the case with the current implementation also) but should be fixed in 3.6.0.
* hiding the tray icon [can cause a crash in libdbusmenu](https://gitlab.archlinux.org/archlinux/packaging/packages/libdbusmenu/-/merge_requests/1) (which current non-SNI implementation in SDL can hit) - it seems to only happen on repeated triggering of show/hide, hopefully, so shouldn't be a large issue
* pulling now will break the scroll-to-change-volume feature - which seemingly only works on GNOME anyways at the moment
* accessibility strings will be removed but I have no idea if anyone even uses it
* This should hopefully implement the macOS tray which is currently broken

PR #1773, opened 2025-08-08
Bump Windows ffmpeg to 7.1.1

C0rn3j:
We still ship an ancient 5.x version at the moment.

Guts out #1598 to get the non-controversial changes from it.

This stops hardcoding the total filesize (except the initial prompt and fail of getting the size from the headers).

Stops hardcoding the exit to a certain filesize.


C0rn3j:
Done and tested, seems to work great, we do not currently have a mechanism to update existing 5.x installations, but that can be added later if desired.

PR #1801, opened 2025-08-24
Reduce complexity of main()

C0rn3j:
main currently has way too much logic, split it into separate functions so we can have LSPs analyze it.

See https://github.com/microsoft/pyright/discussions/10838 for why this is necessary.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -5618,6 +5618,7 @@ class Tauon:
 		self.bag                          = bag
 		self.mpt                          = bag.mpt
 		self.gme                          = bag.gme
+		self.dev_mode                     = bag.dev_mode
 		self.renderer                     = bag.renderer
 		self.ddt                          = TDraw(bag.renderer)
 		self.fonts                        = bag.fonts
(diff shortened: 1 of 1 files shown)


Taiko2k:
Seems like a futile effort to me. Main here is massive, you’d be splitting off code into functions forever. It doesn’t reduce the “complexity” either, it makes it more complex.

I feel like efforts need to be pushed toward bug fixes and stability. Because ultimately that’s what ends users see.

C0rn3j:
> Main here is massive, you’d be splitting off code into functions forever

The problematic parts are the logic branching ones with if's/try/catches, etc, so it's not all that gloomy of an outlook.

I already split 1.7k lines here, and those 1.7k lines here now get analyzed properly.

>  It doesn’t reduce the “complexity” either, it makes it more complex.

From one point of view it does, I meant from the complexity of code when looked at by an LSP.

Note the rest of the codebase seems to work fine, it's just main() that's an issue here.

> I feel like efforts need to be pushed toward bug fixes and stability. Because ultimately that’s what ends users see.

That's actually the end goal, it's much easier to spot a mistake when you get warned about it - i.e. [this](https://github.com/Taiko2k/Tauon/commit/7d41e99c41bdbc608d50db1e3c89f08a61a46905) was found by reading the logs from the tooling, and in the past I've fixed multiple of bugs like these already, so I find it worthwhile to get the LSP functioning across the entire codebase.

C0rn3j:
Let's get the noncontroversial changes in first - I've removed splitting up of the functions - leaving only commented out skeletons around - keeping the refactors that were necessary to make it happen, which we want in either way.

PR #1803, opened 2025-08-25
Fix master_library being loaded wrong

C0rn3j:
It seems we were loading `master_library` into `master_library` but pctl was initializing with it from Bag, which would be empty...


C0rn3j:
Superseded by #1801

Issue #1763, opened 2025-08-04
Lyrics search: genius returns part of the blurb at the start of the lyrics


rexendevar:
If Genius finds lyrics, it will return them with unnecessary data at the top

<img width="316" height="130" alt="Image" src="https://github.com/user-attachments/assets/e1f89726-b3d5-4044-a80f-3942f72b9611" />

This song should have remained blank.

rexendevar:
> submitted a pr that fixes this

If you want Github to recognize it, you can add something to the pr that says "resolves #1763" or something like that 👍

Issue #123, opened 2019-05-26
Various suggestions


StrangeSeasOfThought:
- album artist
- composer (also missing in track info)
- comment

-------

There's some weird behavior in regards to the date tag. All my music is tagged via the following date format: year-month-day (yyyy-mm-dd). Lots of albums only display year (yyyy) in Tauon, several display year-day-month (yyyy-dd-mm), while some correctly display year-month-day (yyyy-mm-dd).

Taiko2k:
- ***album artist:*** Sure, I can add this.
 - ***composer:*** I can look at adding this to the *columns* view, though I'm not sure where it should go in the track info box, its pretty crammed as it is.
- ***comment**:* Hmm... I never thought of that as being useful to see in column view.

As for the date tags... in most places (like the track info box), Tauon displays it exactly as it is formatted in the tag. Maybe theirs a correlation of behavior between different file types? We might need to look more closely at a specific file to see whats going on exactly.

StrangeSeasOfThought:
![playlist](https://user-images.githubusercontent.com/43582772/58398729-41bda080-8056-11e9-8325-9d422cac6cd4.png)
![be1](https://user-images.githubusercontent.com/43582772/58398733-45512780-8056-11e9-98ee-91f45b4c9772.png)
![be2](https://user-images.githubusercontent.com/43582772/58398741-484c1800-8056-11e9-945c-ef9401eca125.png)

I noticed that dates with UTF16 text encoding are being displayed correctly, while dates with a ISO-8859-1 text encoding aren't.

Taiko2k:
What file format/codec are these?

StrangeSeasOfThought:
All are mp3.

Taiko2k:
Yeah I'm not sure whats going on there exactly... I actually use a 3rd party library for reading ID3 tags, so I don't think it's something I can easily fix. I guess the answer is to re-tag all your music in UTF16? 

**\*Taiko2 runs away\***

StrangeSeasOfThought:
> I guess the answer is to re-tag all your music in UTF16?

Yeah, or UTF8. Already done. Everything's fine now.

StrangeSeasOfThought:
> * _**composer:**_ I can look at adding this to the _columns_ view, though I'm not sure where it should go in the track info box, its pretty crammed as it is.

How about right below Album Artist, same row as Duration?

> * _**comment**:_ Hmm... I never thought of that as being useful to see in column view.
[...]

PR #1809, opened 2025-08-29
fix genius lyrics scrape picking up artifacts like "3 Contributors"

0w0v:
fixes https://github.com/Taiko2k/Tauon/issues/1763

removes stuff like
"3 Contributors"
"[Song Name] Lyrics"
"Embed" buttons
"See [Artist] Live"
Ticket advertisements
Related content suggestions


C0rn3j:
`fixes issue #123` won't work, `fixes #123` will, I edited your post

Taiko2k:
Do we want to strip empty lines, would that not remove paragraph breaks?

PR #1810, opened 2025-08-29
the first post-merge bug has been found

rexendevar:
do u want me to do these all in one branch or in a new branch for each bug?


C0rn3j:
It'd be nice to have the commit messages reflect what they actually fix.

You'll need a new branch after each merge, unless you want to rebase it against master each time.

rexendevar:
> It'd be nice to have the commit messages reflect what they actually fix.
> 
> You'll need a new branch after each merge, unless you want to rebase it against master each time.

will do in the future, now that it's not just me keeping track of my own issues in my notes app. lol

PR #1819, opened 2025-09-02
fix edge case which caused infinite failed imports

rexendevar:


Taiko2k:
Oh I went and fixed this one myself

PR #1820, opened 2025-09-05
fix up a couple error messages

rexendevar:
drag and drop import files & failed lrclib upload

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -40779,17 +40779,28 @@ def worker1(tauon: Tauon) -> None:
 					add_file(filepath[0])
 					logging.info(f"will import {filepath[0]}")
 
-	def add_file(path, force_scan: bool = False, show_errors: bool = False) -> int | None:
-		"""Import playlist from filepath""" # TODO (Flynn): add visible errors for bad playlist imports
-		# bm.get("add file start")
+	def add_file(path: str, force_scan: bool = False, show_errors: bool = False) -> int | None:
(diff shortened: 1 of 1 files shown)


src/tauon/t_modules/t_main.py
```diff
@@ -40779,18 +40779,37 @@ def pl_folder_autoscan() -> None:
...
+		except PermissionError:
+			logging.error("Bad permissions to import")
+			if show_errors:
+				tauon.show_message(
+					_("Bad permissions"),
+					_("We don't have permissions to access {file}").format(file=path),
+					mode="error"
+				)
+			return 0
+		except:
```
C0rn3j, line 40805: https://docs.astral.sh/ruff/rules/bare-except/

Both of these need to be `except Exception`

src/tauon/t_modules/t_main.py
```diff
@@ -40779,18 +40779,37 @@ def pl_folder_autoscan() -> None:
...
+			logging.error("Bad permissions to import")
+			if show_errors:
+				tauon.show_message(
+					_("Bad permissions"),
+					_("We don't have permissions to access {file}").format(file=path),
+					mode="error"
+				)
 			return 0
+		except Exception:
+			logging.error("Can't import for some reason")
```
C0rn3j, line 40806: This should be logging.exception in the generic case, so the exception gets logged fully.

If we're catching a specific one, there's no need for a full stack trace since we already know what's going on, but the rest we do want logs on

PR #1827, opened 2025-09-11
fixing a couple issues with saving lyrics

rexendevar:
- [x] bug that removes random tracks from playlists when saving lyrics
original pr text:
I'm still not sure whether the tracks were being deleted from playlists or from the whole database or what. Stopping the currently playing song before writing to it might fix it. If I still get the same bug, I'll attack it again by replacing play() with a modified version of play_target() that should fully reopen the file.
 - [x] add option for automatically saving "simple" lyrics edits (clear, search etc rather than from the editor) rather than mandating it
 - [ ] add waitlist to autosave simple lyrics edits rather than block stopping playback to do so
 - [x] avoid running track end behavior on manual skip
 - [x] "repeat track" as new default track end behavior

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -7759,6 +7759,10 @@ class Tauon:
 				lrc.write( lyrics )
 				logging.info(f"Edited the LRC file for {track.artist} - {track.title}")
 				return True
+		# fully stop and resume track to prevent severe bug when simultaneously reading and modifying
+		stop = track.index == self.pctl.track_queue[self.pctl.queue_step]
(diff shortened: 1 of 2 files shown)


Taiko2k:
I'd like to get *make saving lyrics to files optional* into master. Could you un-draft this PR?

rexendevar:
> I'd like to get _make saving lyrics to files optional_ into master. Could you un-draft this PR?

yes, but be careful! with the current state of things, lyrics changes that save to the disk will force stop playback before saving. this INCLUDES simple changes. i don't have the drive right now to implement a waitlist system so it's up to you

PR #1888, opened 2025-11-09
new column for file change time

rexendevar:
Lets the user see when a file was last changed. Not going to immediately merge this one cause I want a second opinion on its importance/code quality


Taiko2k:
Seems OK to me. Just to note this value is only updated on first scan / tag rescan, it still can be useful though.

PR #1902, opened 2025-12-01
Rename menu item 'Reload All Folders' to 'Rescan All Folders'

Peu255:
This PR makes a very small UI text correction:
“Reload All Folders” → “Rescan All Folders”

The main module `t_main.py` has a different string from `messages.pot`...

`t_main.py`
Captura de tela de 2025-12-01 02-14-19

`messages.pot`
Captura de tela de 2025-12-01 02-07-08

resulting in a bug to translations
Captura de tela de 2025-12-01 02-08-36 (pt-br example)

No logic or UI structure was modified, only the display text.


C0rn3j:
We should do this the opposite way - fix the translation file to reflect the current code.

The way to do that is to run `update_translations.py`.

I've sent the update in #1903 just now, thanks for the PR though!

PR #1926, opened 2025-12-18
Support 4-channel ogg

C0rn3j:
This is entirely vibecoded - it works with the provided 4-channel file on Discord.

Without this patch, the file plays at seemingly 0.5x speed.

### DIFF
--- a/src/phazor/phazor.c
+++ b/src/phazor/phazor.c
@@ -854,6 +854,10 @@ void read_to_buffer_16in32_fs(int32_t src[], int n_samples) {
 	resample_to_buffer(f);
 }
 
+static inline float s16_to_float(const unsigned char *p) {
+	return ((int16_t)((p[1] << 8) | p[0])) / 32768.0f;
+}
+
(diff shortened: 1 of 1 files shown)


Taiko2k:
~~For some tidier code we should probably just use the decoders built in function for this.~~

> There are special functions, [op_read_stereo()](https://opus-codec.org/docs/opusfile_api-0.7/group__stream__decoding.html#gaaca17ad3fd6430dc8fd43308149428eb) and [op_read_float_stereo()](https://opus-codec.org/docs/opusfile_api-0.7/group__stream__decoding.html#ga9736f96563500c0978f56f0fd6bdad83), which always output two channels, to simplify applications which do not wish to handle multichannel audio. These downmix multichannel files to two channels, so they can always return samples in the same format for every link in a chained file.

~~I use those for OPUS, I must have forgotten to do that for Vorbis.~~

C0rn3j:
Feel free to either rework or close this

Taiko2k:
Oh only OPUS has the downmix option not VORBIS, scratch my last post, ill go with your rework

PR #1931, opened 2025-12-20
Update t_main.py

anshag2007:
Fixed a fatal error which arose while exporting albums as CSV, not being able to concat a string and the filepath (fixed by explicitly typecasting the filepath) in line 9137


C0rn3j:
Thanks for the PR!

PR #1937, opened 2025-12-24
fix: fix TIDAL integration (renamed property)

vale981:
(this fixes the issue on nixos. maybe you wan't to pin the tidalapi version)


C0rn3j:
Thanks for the PR!

PR #1948, opened 2026-01-02
Warn when encountering non-standard FMPS tag

C0rn3j:
We've had an user convert from another player by writing AI-assisted code which used non-standard values for the ratings, check this on our side so we don't save a value like "600" to a rating field that's' supposed to be from 0 to 10.


Taiko2k:
This PR just warns though, don't we want to skip if the value is out of range?

C0rn3j:
That's probably a good idea, I'll rework it later

PR #1973, opened 2026-01-14
Implement Subsonic Favorite Synchronization

Micromot20:
My first pull request to an actual project.

This Implements sending new favs/unfavs to the Airsonic Server and also retrieves Favs from the server on a library download/refresh


src/tauon/t_modules/t_subsonic.py
```diff
@@ -0,0 +1,388 @@
...
+				nt.is_network = True
+
+				rating = 0
+				if "userRating" in song:
+					rating = int(song["userRating"])
+				
+				if nt.url_key in liked_tracks:
+					liked_track_ids.append((
+						nt
+					))
```
C0rn3j: You're adding the entire track object here instead of just an id, that doesn't seem intended?

src/tauon/t_modules/t_subsonic.py
```diff
@@ -0,0 +1,389 @@
...
+		
+		try:
+			a = self.r("getStarred2")
+		except Exception:
+			logging.exception("Error connecting to Airsonic server")
+			self.show_message(_("Error connecting to Airsonic server"), mode="error")
+			return []
+		b = a["subsonic-response"]["starred2"]
+
+		liked_tracks: list[tuple[str]] = []
```
Taiko2k: Why the `tuple` here? Why not simplify it? But also see below.

src/tauon/t_modules/t_subsonic.py
```diff
@@ -0,0 +1,389 @@
...
+					))
+
+				playlist.append(nt.index)
+
+				if self.star_store.get_rating(nt.index) == 0 and rating == 0:
+					pass
+				else:
+					self.star_store.set_rating(nt.index, rating * 2)
+
+		def set_favs(d:list[tuple[str]]) -> None:
```
Taiko2k: Tauon track IDs (`track_object.index`) are not the same as Airsonic IDs. `star_store` takes Tauon track IDs.

Could you also change `tr` to something like `track_id` here to make it more clear. I used `tr` elsewhere to refer to a track object.

Taiko2k (changes requested):

Taiko2k:
merge conflict, can you try rebase?

C0rn3j:
I've rebased it, fixed the imports and touched it up a little.

PR #1988, opened 2026-01-21
Switch to native SDL taskbar progress

Taiko2k:


C0rn3j:
Should also be removed from here

https://github.com/Taiko2k/Tauon/blob/49d5acb2498f68041a0441ea680bb826685b6a57/.github/workflows/build_and_release.yaml#L394

https://github.com/Taiko2k/Tauon/blob/49d5acb2498f68041a0441ea680bb826685b6a57/windows.spec#L53

and probably tested on macOS, since we're using it there and it's not supported - does it just silently fail and things keep working?

Should also inform in the changelog that this is now supported on Linux

C0rn3j:
```
  File "/tmp/Tauon/src/tauon/t_modules/t_main.py", line 1703
    self.updated_state = 0
                          ^
IndentationError: unindent does not match any outer indentation level
```

Looks like parts mix tabs and parts spaces

C0rn3j:
That works... though if you have more Tauon instances open it seems to apply to them all:

image

EDIT: This seems to go deeper, the right click menu is also wrong, they're both for the system instance and not my testing instance that uses this branch... so if I play music from the testing instance I can't stop it by right click, since it shows (and controls) state of the system instance

image

Taiko2k:
@C0rn3j Though FYI in regards to the action buttons, Tauon uses an HTTP server for inter process communication, but only the first instance will be bound to the port, other instances wont be controllable. (Though there's double up there in the menu so I assume Plasma is adding MPRIS controls there too (the ones with the icons), which might also generate confusion between instances.)

Not sure why the progress is being shared though.

PR #1993, opened 2026-01-27
Add macos desktop media control

Taiko2k:
Its horrible... its ugly... and it... works?! We AI slop now. But by god though I would never want to code this by hand. Thank you ChatGPT 5.2, you so smart.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -39166,7 +39166,7 @@ class Bag:
 	default_playlist:        list[int] = field(default_factory=list[int])
 	sm:                      CDLL | None = None
 	nowplaying:              bool = False
-	nowplaying_helper:       object | None = None
+	nowplaying_helper:       MacNowPlayingHelper | None = None
 	song_notification:       None = None
 	active_playlist_viewing: int = 0
 	active_playlist_playing: int = 0


src/tauon/t_modules/t_main.py
```diff
@@ -39134,6 +39165,8 @@ class Bag:
 	loaded_asset_dc:         dict[str, WhiteModImageAsset | LoadImageAsset]
 	default_playlist:        list[int] = field(default_factory=list[int])
 	sm:                      CDLL | None = None
+	nowplaying:              bool = False
+	nowplaying_helper:       object | None = None
```
C0rn3j: I think this should be of type `MacNowPlayingHelper | None`?

C0rn3j:
If this is something we wanna ship, it'd be best to add it to https://github.com/Taiko2k/Tauon/blob/5e9d8031737e2b74c1930fad1743b2d162bf7bd6/.github/workflows/build_extras.yaml if that's possible

If you decide to add it there, you can test the workflow via https://github.com/Taiko2k/Tauon/actions/workflows/build_extras.yaml since it does not run automatically

.github/workflows/build_and_release.yaml
```diff
@@ -268,6 +272,10 @@ jobs:
         env:
           DYLD_LIBRARY_PATH: "/opt/homebrew/lib"
 
+      - name: Ensure Now Playing helper is executable
+        run: |
+          chmod +x "dist/TauonMusicBox.app/Contents/Resources/lib/TauonNowPlaying.app/Contents/MacOS/TauonNowPlaying" || true
```
C0rn3j: Why || true? That'll mask an error if any, chmod does not error if the executable permission is already given

PR #2016, opened 2026-02-15
Enumize QueueType and fix a queue bug

C0rn3j:
The bug fix requires special attention, I am unsure if it is fixed correctly as it is still missing the playlist ID.


src/tauon/t_modules/t_extra.py
```diff
@@ -41,6 +41,7 @@
 from typing import TYPE_CHECKING
 
 from gi.repository import GLib
+from tauon.t_modules.t_enum import QueueType
```
Taiko2k: Should be `t_enums`?

C0rn3j: Whoops, it should, fixed

PR #2025, opened 2026-02-19
Fix importing synced lyrics from FLAC and OPUS files

C0rn3j:
@rexendevar particularly interested if https://github.com/Taiko2k/Tauon/pull/2025/changes/8d1ccaabde62f39aa20b58a120401abdc0e13ccc is problematic for some unforeseen reason


rexendevar:
It's not a problem, I assumed that if the lrc file was gone then we would
remove the associated tag but if not then there's no issue removing it

PR #2028, opened 2026-02-21
Re-enable Open keymap file button

C0rn3j:
I think the input.txt button is a good idea, however I am unable to figure out how to properly space the buttons, the way I did it the spacing is slightly off.

image

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -24432,26 +24432,6 @@ class Over:
 		if self.func_page == 0:
 			y += 23 * gui.scale
 
-			self.toggle_square(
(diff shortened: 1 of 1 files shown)


Taiko2k:
Reload needs to go after config, the others can go on their own line

C0rn3j:
How's this? I moved `Enable Listen Along` to the last page to get space for it.

I wonder if I shouldn't position the X of reload to the Open Keymap File button though.

image

Taiko2k:
If anything the reload needs to be visually closer to config button as thats what its associated with.

C0rn3j:
Moved it closer

image

PR #2065, opened 2026-03-11
fix(config): correct typo in locale comment (zh_CH -> zh_CN)

dotdoge2819:
This PR fixes a typo in the configuration template where the ISO code for Simplified Chinese was incorrectly listed as zh_CH. The correct code is zh_CN.

Why?
New users following the comment in the generated Tauon.conf would be misled, causing the Chinese interface to fail to load.

Changes
Modified src/tauon/t_modules/t_main.py to update the example string.

Verification
Manually verified that zh_CN correctly enables the Simplified Chinese UI.


C0rn3j:
Thanks for the PR!

PR #2071, opened 2026-03-17
Test updated macOS icon

Taiko2k:
Testing updated macOS icon


C0rn3j:
> runs-on: macos-26

Afaik we should keep this runner as old as possible, so the resulting binary is compatible with older macOS releases, bumping it in CI bumps the minimum requirement for users too.

PR #2080, opened 2026-03-20
Fix colours being overriden in light mode

### LINKED ISSUE #2079: Impossible to change synced lyrics color
Hey! I'm been making a light theme for Tauon. Pretty much everything is fine, but I found something that I'm unable to change the color even if I should : `active lyric`. My theme has these lines
```
80,80,80    		lyrics text
30,30,30    		active lyric
```
where the first line changes correctly the text color of non-active lyrics, but whathever I put for the values of active lyric I get the same purple color.... Thanks!

EDIT: The first line is a working example, the second line is the one not working.

C0rn3j:
Fixes #2079, though the forced re-colour I removed seems intended to be there?

I can rework it to only exclude the lyric colours if necessary.


Taiko2k:
I committed a fix for this

PR #2088, opened 2026-04-05
fix: normalize Jellyfin server URL to prevent 404 from trailing slash

### LINKED ISSUE #1076: Jellyfin auth error 404 - please normalize URLs
First time using Tauon, freshly installed from AUR. I tried to setup my own Jellyfin home server, and I get the following message:

> Jellyfin auth error
> 404

I scratched my head, I double-checked the credentials and the URL. It works fine in the browser.

## The bug

After a lot of search, I found this line: https://github.com/Taiko2k/TauonMusicBox/blob/71cd26cf20188ded5a3386dc58dea1d3892df319/t_modules/t_jellyfin.py#L64-L65

It blindly concatenates strings to form the URL.

That means:

* `http://192.168.xx.xx:8096` works fine
* `http://192.168.xx.xx:8096/` fails

## Proposed solution

I have a two-part solution:

### Part 1: form input normalization

When the user types the URL, some basic normalization should be performed:

* Strip away known URL suffixes to generate URL candidates.
    * `http://192.168.xx.xx:8096/web/index.html` → `http://192.168.xx.xx:8096/`
    * `http://192.168.xx.xx:8096/web/` → `http://192.168.xx.xx:8096/`
    * `http://192.168.xx.xx:8096/mymusic/` stays as is. For unknown suffixes, it is likely the Jellyfin is being served as a subpath or using some kind of proxy.
    * `http://192.168.xx.xx:8096/mymusic/index.html` → `http://192.168.xx.xx:8096/mymusic/`
    * `http://192.168.xx.xx:8096/mymusic/web/index.html` → `http://192.168.xx.xx:8096/mymusic/`
* Add `http://` and `https://` prefixes, if the protocol is not specified.
* Add the common Jellyfin port(s) (e.g. `:8096`) if the port is not specified.
* Collect all these candidate URLs by applying several rules and combining the rules.
* Check if each of the normalized candidate URLs is a valid Jellyfin instance.
* Update the form with the normalized URL that is valid.

### Part 2: sane URL formation

The code that forms URLs should using [urllib.parse.urljoin()](https://docs.python.org/3.11/library/urllib.parse.html#urllib.parse.urljoin) instead of blindly concatenating strings.

tmchow:
A trailing slash in the Jellyfin server URL (e.g. `http://192.168.xx.xx:8096/`) caused 404 errors because all API paths were built via f-string concatenation, producing double slashes like `http://server//Users/AuthenticateByName`.

Added a `server_url` property that strips trailing slashes from `prefs.jelly_server_url`, and replaced all 12 direct references in the file with the new property.

Fixes #1076

This contribution was developed with AI assistance (Codex).

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/src/tauon/t_modules/t_jellyfin.py
+++ b/src/tauon/t_modules/t_jellyfin.py
@@ -62,6 +62,11 @@ class Jellyfin:
 		self.session_last_item: dict[str, list[str] | bool | int | str] | None = None
 		self.playlists = []
 
+	@property
+	def server_url(self) -> str:
+		"""Return the Jellyfin server URL with trailing slashes stripped."""
+		return self.prefs.jelly_server_url.rstrip("/")
+
 	def _get_jellyfin_auth(self) -> str:
 		auth_str = f"MediaBrowser Client={self.tauon.t_title}, Device={self.tauon.device}, DeviceId=-, Version={self.tauon.t_version}"
 		if self.accessToken:
@@ -71,7 +76,7 @@ class Jellyfin:
 	def _authenticate(self, debug: bool = False) -> None:
 		username = self.prefs.jelly_username
 		password = self.prefs.jelly_password
-		server = self.prefs.jelly_server_url
+		server = self.server_url
 
 		try:
 			response = requests.post(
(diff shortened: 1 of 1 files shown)


tmchow:
thanks for the merge @Taiko2k!

Issue #2111, opened 2026-04-22
Lyrics don't work when sourcing from Airsonic/Subsonic/Navidrome.


homemdesgraca:
**Desktop**
- OS: CachyOS x86_64 (Arch Linux)
- OS version: Non-applicable (Kernel: Linux 6.19.12-1-cachyos)
- Tauon Version: 9.1.3
- Installation method: OS packager (pacman)

**Describe the bug**
When sourcing songs from Airsonic (including Subsonic/Navidrome), lyrics aren't visible, even when fully available by the provider. 

**To Reproduce**

1. Menu -> Settings -> Accounts -> Airsonic
2. Enter credentials and import music to playlist.
3. Go to generated playlist and play any song that has confirmed lyrics available (static or synced). 
4. Lyrics are not displayed.

**Expected behavior**
The lyrics should be available.
When sourcing DIRECTLY from the files (without Airsonic as a middleman), it works fine.

For comparison, when using Symfonium (an Android music player) sourcing from Subsonic (specifically Navidrome), it fetches lyrics just fine.

**Screenshots/Videos**

<img width="1908" height="1038" alt="Song sourced locally" src="https://github.com/user-attachments/assets/d422f83f-fa03-4d48-99e3-ec3df55dac98" />
Song sourced locally from file.

-----------

<img width="1908" height="1038" alt="Song sourced through Airsonic" src="https://github.com/user-attachments/assets/d6cb20a3-d8ec-48d0-b72d-ca1cb3c8e10c" />
Same song sourced through Airsonic.

Taiko2k:
Committed possible fix (not tested) (rescan not needed, it should fetch them on playback now)

homemdesgraca:
> Committed possible fix (not tested) (rescan not needed, it should fetch them on playback now)

<strike>Hi Taiko! Thank you for replying so quickly :]. I just built Tauon with the latest commit but it still fails to fetch the lyrics.

<img width="1059" height="1034" alt="Image" src="https://github.com/user-attachments/assets/9ff30148-092e-4943-aeba-5710a8989031" />

No log about fetching lyrics on the console:

<details>
<summary>Logs after changing tracks</summary>
2026-04-23 14:52:53 [ INFO  ] [ t_phazor ] Music Baby -> Flash in the Pan
2026-04-23 14:52:53 [ INFO  ] [ t_phazor ]  --- length: 0.0
[...]

PR #2114, opened 2026-04-25
Prefer using synced lyrics on MPRIS when available and refresh MPRIS after Subsonic lyrics fetching logic.

homemdesgraca:
Fixes a related problem with #2111.

Two small changes to fix synced lyrics not appearing through MPRIS for Subsonic streams.

For local files, synced (LRC) lyrics are read from embedded tags or .lrc files and stored in track.synced, so MPRIS shows them correctly. But for Subsonic/Navidrome, the structured lyrics API returns plain text and LRC content as separate values. Tauon was putting them into track.lyrics (plain) and track.synced (LRC), respectively. Since MPRIS only reads track.lyrics, synced lyrics were lost.

Fix 1: In t_dbus.py, prefer track.synced over track.lyrics when publishing MPRIS metadata, so timestamped lyrics show regardless of source. This is a global fix that applies to any provider populating track.synced.

Fix 2: In t_subsonic.py, call notify_update() after the async lyrics fetch completes, so MPRIS consumers pick up newly fetched lyrics immediately, without weird interactions when switching tracks too fast.

### DIFF
--- a/src/tauon/t_modules/t_dbus.py
+++ b/src/tauon/t_modules/t_dbus.py
@@ -129,7 +129,7 @@ class MPRIS(dbus.service.Object):
 				"xesam:albumArtist": dbus.Array([track.album_artist]),
 				"xesam:artist": dbus.Array([track.artist]),
 				"xesam:title": track.title,
-				"xesam:asText": track.lyrics,
+				"xesam:asText": track.synced if track.synced else track.lyrics,
 				"xesam:autoRating": star_count2(self.tauon.star_store.get(track.index)),
 				"xesam:composer": dbus.Array([track.composer]),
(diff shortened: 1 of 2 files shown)


Taiko2k:
Not sure if exposing synced lyrics over MPRIS will cause issues for other consumers. Im merging this on the basis it may be reverted if its apparent its an issue.

homemdesgraca:
> Not sure if exposing synced lyrics over MPRIS will cause issues for other consumers. Im merging this on the basis it may be reverted if its apparent its an issue.

Yeah I thought about that. But I assumed that most MPRIS consumers expect that behavior in some way. Most projects that uses lyric info provided by MPRIS use it for displaying lyrics, so exposing only static broke them.

Issue #2156, opened 2026-05-20
Watch theme file changes


XLhinares:
**Problem description:**
I use dynamic theming to match wallpaper colors whenever I can, and made a script to generate a theme `wallpaper.ttheme` on wallpaper change. However, to apply the changes to Tauon, I need to press F10.

**Preferred solution:**
Ideally, I wish Tauon would watch for changes to the current ttheme file, and reload theme when needed.

**Alternatives solutions:**
An alternative solution could be to make an API so Tauon can be programmatically told to reload the theme. Maybe something like a `tauon --reload-theme` command.

**Additional context:**
No extra context, this is really just a quality of life request tbh
Thank you for the great app!

XLhinares:
My problem was solved by 2bdf64fd50794caa355cd5344b06ebd98e175cab which added the `tauon --reload-theme` command argument.

PR #2158, opened 2026-05-20
Add support for reloading the theme from command line

XLhinares:
This solves the problem I brought up in the issue #2156

## What it does:

This PR allows the user to run `tauon --reload-theme` to make tauon (re)load its theme.

## Why:

I use a dynamically generated `wallpaper.ttheme` file that is overwritten on wallpaper change.
Having this feature allows my theme-generation script to also tell tauon to reload its theme (instead of having to manually press F10).

## Note:

- In my original issue, I suggested watching theme file change; but after a quick code review, I figured that this solution fits the existing code base a lot better, so I went with that.
- I noticed lines like `elif path == "/api1/<argument>":` in `src/tauon/t_modules_t_webserve.py` (`doGet` function) that look a lot like the ones I added, but did not touch them because (1) I am not familiar enough with the code and (2) figured there's no use in adding support for remotely reloading the theme.

### DIFF
--- a/src/tauon/t_modules/t_webserve.py
+++ b/src/tauon/t_modules/t_webserve.py
@@ -1093,6 +1093,8 @@ def controller(tauon: Tauon) -> None:
 				tauon.toggle_random()
 			if path == "/repeat":
 				tauon.toggle_repeat()
+			if path == "/reloadtheme":
+				tauon.load_theme()
 			if path.startswith("/open/"):
 				rest = path[6:]
 				try:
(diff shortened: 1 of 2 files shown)


src/tauon/t_modules/t_webserve.py
C0rn3j, line 1076: You are adding a path which already exists.

XLhinares: My bad, these lines totally do what I wanted and I somehow missed them 😅
Thanks for letting me know!

XLhinares: I removed my changes to the `t_webserve.py` file

C0rn3j (approved):
LGTM

PR #2172, opened 2026-05-30
Specify Plex music library

hellbound22:


Taiko2k:
I'm thinking maybe this should default to "Music", otherwise will users know they need to specify that? (If that's how it works, I'm not too familiar)

hellbound22:
The default library names on plex change according to what the language of the server is set at the initial setup. I've seen some apps with plex integration that are able to fetch a list every music library independently, maybe that's the correct way to do it? Else we would need to rely on a file that has the default music library name for every single language.

Imo it's either list all the libraries in a drop-down or set a default for the input box

Issue #1842, opened 2025-09-29
Tauon crashes


Solarunit:
**Desktop**
- OS:  Solus
- Tauon Version:  8.1.4
- Installation method: Flatpak

**Describe the bug**

Multiple crashes in the log (probably when listening radio)

**Tauon log**

```python
 Traceback (most recent call last):
  File "/app/bin/src/tauon/__main__.py", line 464, in <module>
    main()
  File "/app/bin/src/tauon/__main__.py", line 460, in main
    t_main(holder)
  File "/app/bin/src/tauon/t_modules/t_main.py", line 44257, in main
    tauon.radio_view.render()
  File "/app/bin/src/tauon/t_modules/t_main.py", line 35047, in render
    self.tauon.toast(_("Added station to: ") + pctl.radio_playlists[pctl.radio_playlist_viewing]["name"])
                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
TypeError: 'RadioPlaylist' object is not subscriptable


```

```python
Traceback (most recent call last):
  File "/app/bin/src/tauon/__main__.py", line 475, in <module>
    main()
  File "/app/bin/src/tauon/__main__.py", line 471, in main
    t_main(holder)
  File "/app/bin/src/tauon/t_modules/t_main.py", line 46038, in main
    radiobox.render()
  File "/app/bin/src/tauon/t_modules/t_main.py", line 31557, in render
    self.draw_list()
  File "/app/bin/src/tauon/t_modules/t_main.py", line 31638, in draw_list
    scroll_distance = self.scroll.scroll("radio box")

AttributeError: 'ScrollBox' object has no attribute 'scroll'. Did you mean: 'coll'?
```

```python
Traceback (most recent call last):
  File "/app/bin/src/tauon/__main__.py", line 475, in <module>
    main()
  File "/app/bin/src/tauon/__main__.py", line 471, in main
    t_main(holder)
  File "/app/bin/src/tauon/t_modules/t_main.py", line 46026, in main
    tauon.rename_track_box.render()
  File "/app/bin/src/tauon/t_modules/t_main.py", line 22038, in render
    if self.rename_files.text != self.prefs.rename_tracks_template and self.draw.button(
                                 ^^^^^^^^^^
AttributeError: 'RenameTrackBox' object has no attribute 'prefs'


```

```python

Traceback (most recent call last):
[...]

Issue #1597, opened 2025-04-28
Radio stops playing after switching stations


GodTGamer:
**Desktop (please complete the following information):**
 - OS: Bazzite Linux and (Windows 11)
 - OS version: 42 - Nvidia Edition
 - Tauon Version: 8.0.1 (Pre-release-Tauon-2025-04-27-7d3da75)
 - Installation method: Flatpak
**Describe the bug**
Radio stops playing after switching to another station. Sometimes it's possible to switch a few times, but mostly it stops after the first switch, and then it's impossible to play radio again. (Music playing is still possible)
Tauon now always plays the loading animation in the play bar if you don't play a song from your local library.

**To Reproduce**
1. Go to 'Menu'
2. Click on 'Radio'
3. Choose a Radio Station
4. Click on 'Play'
5. Click on 'Stop' (optional)
6. Choose another Radio Station
7. Click on 'Play'
8. Sometimes you have to Repeate 5.-7. till its stops

**Expected behavior**
I expected that the Radio switch to another station and plays music.

**Screenshots/Videos**
![Image](https://github.com/user-attachments/assets/7b9c6028-ea0a-44de-9264-8934e82e01aa)
![Image](https://github.com/user-attachments/assets/cc70b08d-6a87-4aba-ab06-358c2052d6ce)

**Tauon log**

[domsch@bazzite domsch]$ flatpak run com.github.taiko2k.tauonmb
2025-04-28 20:38:45 [ INFO  ] [ __main__ ] Tauon v8.0.1
2025-04-28 20:38:45 [ INFO  ] [ __main__ ] Copyright 2015-2025 Taiko2k captain.gxj@gmail.com

2025-04-28 20:38:46 [ INFO  ] [ __main__ ] SDL window system: wayland
2025-04-28 20:38:46 [ INFO  ] [  t_draw  ] Found jxlpy for JPEG XL support
2025-04-28 20:38:46 [ INFO  ] [  t_main  ] Window size: [1400, 750]
2025-04-28 20:38:46 [ INFO  ] [  t_main  ] Detected running as Flatpak
2025-04-28 20:38:46 [ INFO  ] [  t_main  ] Platform: linux
2025-04-28 20:38:46 [ INFO  ] [  t_main  ] Installed as Flatpak
2025-04-28 20:38:46 [ INFO  ] [  t_main  ] Install directory:         /app/bin/src/tauon
[...]

PR #2178, opened 2026-06-07
Fix radio streaming: BrokenPipeError crashes and websocket connection errors on station switching

### LINKED ISSUE #943: Radio stops playing
Radio stops playing very often, here is exception:

```
Send heatbeat
{"op":10}
Send heatbeat
{"op":10}
Websocket closed
Abort stream connection
Exit feeder
Exception in thread Thread-32 (send_heartbeat):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/app/bin/t_modules/t_main.py", line 34962, in send_heartbeat
    ws.send("{\"op\":9}")
  File "/app/lib/python3.10/site-packages/websocket/_app.py", line 204, in send
    raise WebSocketConnectionClosedException(
websocket._exceptions.WebSocketConnectionClosedException: Connection is already closed.
```

PS: listen.moe radio

Copernicium282:
Fixes #943 , #1597 and #1842 

### Notes

**Issue #943 & #1597** - Radio crashes when switching stations due to `BrokenPipeError` in the feed thread. The `feed()` function in `t_stream.py:pump()` catches exceptions, logs them via `logging.exception()`, then re-raises, causing the daemon thread to crash instead of exiting gracefully.

**Issue #1842** - `WebSocketConnectionClosedException` crashes when switching away from listen.moe stations. The `send_heartbeat()` and op=0 handlers call `ws.send()` without handling closed connections. Additionally, when listen.moe sends track data with `image: null` in album info, the URL construction crashes with `TypeError`.

### Fixes applied:

**t_stream.py:365-369** - Removed `raise` after `logging.exception("Feed not running!")`. The finally block already sets `feed_running = False` and closes the pipe. The thread now exits gracefully on broken pipe errors during station switching.

**t_main.py:35125-35148** - Wrapped `ws.send()` calls in try-except blocks to handle `WebSocketConnectionClosedException` gracefully. Logs "Websocket already closed" when the connection is closed during station switching.

**t_main.py:35165-35172** - Added null check for album image before constructing cover URL. Uses `.get("image")` and only fetches if image exists, preventing TypeError when listen.moe sends null images.

### Testing

- py_compile passes for both
- Runtime: Play listen.moe radio, rapidly switch to SomaFM stations, verify clean transitions with "Websocket already closed" and "Feed not running!" log messages instead of crashes, here's my log file: 
[tauon_test.log](https://github.com/user-attachments/files/28677954/tauon_test.log)

### Log Notes

Pre-existing unrelated errors in the log file (not from these fixes):
- `KeyError: 'artistbackground'` — fanart lookup failures for artists without fanart
- `ResourceWarning: unclosed file` — decoder process cleanup warnings.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -35125,7 +35125,10 @@ class RadioBox:
 			def send_heartbeat(ws: WebSocketApp) -> None:
 				#logging.info(self.ws_interval)
 				time.sleep(self.ws_interval)
-				ws.send("{\"op\":9}")
+				try:
+					ws.send('{"op":9}')
+				except Exception:
+					logging.info("Websocket already closed")
 				logging.info("Send heartbeat")
 
 			def on_message(ws: WebSocketApp, message: str) -> None:
@@ -35138,7 +35141,10 @@ class RadioBox:
 
 				if d["op"] == 0:
 					self.ws_interval = d["d"]["heartbeat"] / 1000
-					ws.send("{\"op\":9}")
+					try:
+						ws.send('{"op":9}')
+					except Exception:
+						logging.info("Websocket already closed")
 
 				if d["op"] == 1:
 					try:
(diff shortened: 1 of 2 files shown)


Solarunit:
Thanks for fixing those radio related issues! :+1: 

___
BTW I noticed in v10 that Play/Pause shortcut pauses playing radio but when you press Play it would play your local playlist instead of starting playing radio agan. I probably need to create a new issue.

Copernicium282:
> Thanks for fixing those radio related issues! 👍
> 
> BTW I noticed in v10 that Play/Pause shortcut pauses playing radio but when you press Play it would play your local playlist instead of starting playing radio agan. I probably need to create a new issue.

Hey, no prob 😀
if you do create an issue, please do upload the log also, it'll be helpful for me in debugging
I'm currently targeting old unresolved bugs, and may look into it later on.

PR #2187, opened 2026-06-13
System to theme multiple elements at a time

rexendevar:
Theme elements can be grouped together under the hood. The theme editor can now handle every single color that gets saved in a theme (except for tb_line which did nothing).

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -29912,7 +29912,7 @@ class Over:
 			swatch_rect = (row_rect[0] + round(8 * gui.scale), row_rect[1] + round(5 * gui.scale), round(14 * gui.scale), round(14 * gui.scale))
 			ddt.rect(swatch_rect, component_colour)
 			ddt.rect_s(swatch_rect, alpha_blend(ColourRGBA(255, 255, 255, 40), row_border), round(1 * gui.scale))
-			ddt.text((swatch_rect[0] + swatch_rect[2] + round(8 * gui.scale), row_rect[1] + round(4 * gui.scale)), _(label), colours.box_text if active else colours.box_button_text, 12, bg=row_fill, max_w=row_rect[2] - round(44 * gui.scale))
+			ddt.text((swatch_rect[0] + swatch_rect[2] + round(8 * gui.scale), row_rect[1] + round(4 * gui.scale)), _(label), colours.box_text if active else colours.box_button_text_highlight if hover else colours.box_button_text, 12, bg=row_fill, max_w=row_rect[2] - round(44 * gui.scale))
 			row_y += row_step
 
 		right_inner_x = right_rect[0] + pad
@@ -36607,7 +36607,7 @@ class PlaylistBox:
 		visible_scroll_rows = max(0, ((h - top_pad - bottom_pad - self.tab_h) / max(row_step, 1)) + 1)
 		max_scroll = max(len(pctl.multi_playlist) - visible_scroll_rows, 0)
 
-		tab_title_colour = ColourRGBA(230, 230, 230, 255)
+		tab_title_colour = self.colours.tab_text
 
 		bg_lumi = test_lumi(self.colours.playlist_box_background)
 		light_mode = False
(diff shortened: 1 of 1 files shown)


Taiko2k:
But what if the user wants to theme those individually. Should we not have a way to ungroup those like an advanced mode or something?

rexendevar:
I don't really think we should. I'm not at all opposed to some system to help people make advanced themes (maybe a keybind to reload the current theme from the file if we don't have one already, plus an indicator in the theme menu for when a color group actually has multiple different colors assigned) but I think putting in an advanced mode within the Tauon GUI would be too much effort for the benefit. 
- People have been making themes by editing the text files for most of the time the program has existed. I see the theme editor as a way to lower the skill floor, not a way to make something possible that wasn't before.
- It sounds like a lot of effort adding a new mode switch and a new massive tuple only to end up with a list that's a mile and a half long.
- Some colors cannot be displayed at the same time as the theme editor window, e.g. anything relating to the context menu or the mini mode. The theme editor primarily allows for live preview of colors; I'd argue that providing a keybind to reload the theme would be more useful than either extending the list of colors you're editing blind or trying to do some kind of smart UI preview display

rexendevar:
& I'm also not opposed to rethinking the specific color groups I have set up for this. as long as we get color groups I'm happy

Solarunit:
Current Theme Editor doesn't have many components you can set colors for.
Do you propose to limit the list of components by combining several components into a single one?

BTW I can't find how to set color for Playlist panel.

rexendevar:
> Do you propose to limit the list of components by combining several components into a single one?

Yes. The idea is to make theming less of a Task and more something you can do when you're a little bored.

> BTW I can't find how to set color for Playlist panel.

That's one of the colors that's not in the current list, lol. If you're ok with editing the text file until this PR gets figured out & merged, the label should be playlist_box_background

Solarunit:
>Yes. The idea is to make theming less of a Task and more something you can do when you're a little bored.

I don't see the reason of reducing number of components (I count 35 of them, not that much)
There could be an option to show all of them for advanced users, e.g. some toggle next to "Components"


>That's one of the colors that's not in the current list, lol. If you're ok with editing the text file until this PR gets figured out & merged, the label should be playlist_box_background

I never tried to edit theme via config file. First time I tried it in v.10 Theme Editor. 
I like Theme Editor and live preview feature.

PS: Theme Editor is a new killer feature!

rexendevar:
I don't know where you're getting 35. I counted 53 color groups (as set up in my PR) and 86 total themeable colors. And that's excluding colors like `vis_bg` (background color of the visualizer in the top right) which can be customized within the code but don't get saved in the theme file. My angle would be to round up all the colors like that (all the ones that make sense to theme anyway) and get them all themeable, likely by putting them into these bins somehow.

Solarunit:
I opened my favorite theme Cucumber Splash and I see about 35 items.
For Ash theme it's about the same.

Maybe some newer theme has more colors?

rexendevar:
I have a possible UI compromise for advanced mode: for each of my color bins that modify multiple elements, we can provide an expandable dropdown to optionally show the component elements on a bin-by-bin basis. So if a person wants to set their volume bar and their seek bar to separate colors, they can do that without also having to see things like the "status button" or "active lyric"

rexendevar:
> Maybe some newer theme has more colors?

The current theme editor does not show all colors that can be themed. It's completely missing things like the playlist box background or anything to do with the colors of buttons, and doesn't adapt to the theme file it's editing. My PR makes every theme color that currently saves to a file accessible

Solarunit:
So there could be items groups that you can expand. 
What about gobal checkbox to expand all?

Or groups would have some Titles? Then double click on a title could expand/shrink the list

rexendevar:
Groups would have titles, they'd also have their own color selector (which'll be crossed out or something in the list display when its subcomponents are set to different colors), and there'll be a way to expand all (though I lean towards a pair of regular buttons to expand/collapse all). Whole idea is you can either edit a whole group at once or you can go deeper and edit the individual elements.

Solarunit:
Sounds good. I like the idea that user could edit any element

Taiko2k:
> Groups would have titles, they'd also have their own color selector (which'll be crossed out or something in the list display when its subcomponents are set to different colors), and there'll be a way to expand all (though I lean towards a pair of regular buttons to expand/collapse all). Whole idea is you can either edit a whole group at once or you can go deeper and edit the individual elements.

Sounds like a good solution to me.

rexendevar:
Can I merge this and treat the dropdown system as a future feature?

Taiko2k:
I don't like how it locks off certain things from being individually themable, if something is its own variable, it was intended to be themable by itself. I'm not against simplification, but I think a better short term simplification would be to just hide the `box_*` elements behind an advanced toggle.

rexendevar:
I think if you don't consider the group system to be ready, then we should have a separate PR to add all the remaining themeable colors to the list that already exists and then incubate this PR for a bit.

rexendevar:
Again, I should mention it doesn't lock off the individual elements from being themable. It just makes it as hard to individually theme them as all aspects of the theming process have been for most of the program's lifespan.

rexendevar:
@Taiko2k I added the group dropdowns (easier than i expected) it could do with refinements but would you consider it good to merge now?

PR #2191, opened 2026-06-15
feat: Refactored and Enhanced CLI Shell Script

cilginc:
I noticed that the Arch Linux packaged shell script (`/usr/bin/tauon`) was missing some native features, so I decided to fix it. I've restored the missing functionality and added some extra CLI spice to it.

Funnily enough, I was actually looking for a hacky way to reload themes for `matugen` and was even thinking about implementing it myself. What a coincidence that it was already implemented but just undocumented.

**Note**: Updating the GitHub wiki and the website at some point would be great too, as quite a few things seem a bit outdated there.

Let me know if you have any feedback, ideas, or extra features you'd like to see before merging.

### DIFF
--- a/extra/com.github.taiko2k.tauonmb.sh
+++ b/extra/com.github.taiko2k.tauonmb.sh
@@ -6,7 +6,6 @@ GREEN=$(echo -en '\033[00;32m')
 YELLOW=$(echo -en '\033[00;33m')
 BLUE=$(echo -en '\033[00;34m')
 MAGENTA=$(echo -en '\033[00;35m')
-CYAN=$(echo -en '\033[00;36m')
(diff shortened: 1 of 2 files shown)


Taiko2k:
Not sure I like the hacky way of getting the version number there. Maybe just omit the version option to keep it simple

cilginc:
> Not sure I like the hacky way of getting the version number there. Maybe just omit the version option to keep it simple

To be honest, I was a bit on the fence about this too initially. However, since most standard CLI tools include a `--version` flag, I figured it would be a helpful addition for user experience.

From a safety standpoint, when `--version` is called, the worst case scenario getting exit(1), which doesn't affect using the application.

I'm also fine with the simple version though. So i pushed a fix.

PR #2195, opened 2026-06-17
Clear crash log file before logging current crash

rexendevar:


C0rn3j:
This is a bad idea, you can have a problem cause an issue X that writes some malformed data, crash as a consequence, and then each subsequent launch will crash on issue Y trying to process that malformed data.

If you overwrite the file, you won't be able to sanely debug the initial condition X that got you there, you'll just know you're breaking on loading bad data.

That's not a theoretical, we already had that happen, I just can't remember what it was.

I don't think that users will mind some kilobytes of extra taken storage if we don't clear it, but if you want to clear it anyway, I would do it via checking the last modified time of the crash log, and if it's something older than lets say 6-12 months, delete it.

rexendevar:
You bring up a good point. I'm totally willing to make it only do this given a certain file age. That said I think the file age threshold should be more like a couple days. It's not storage space I'm worried about, more the possibility that someone who doesn't understand computers that well is gonna upload 50 pages of old logs when only the most recent one or two will do.

C0rn3j:
> I think the file age threshold should be more like a couple days

I can definitely imagine a user that has an issue, doesn't bother reporting it or solving it, tries actually solving it again weeks/months later and that would then delete the log.

The logs are timestamped so having them spammed at us ain't much of an issue.

rexendevar:
I see what you're saying, but if they don't get the same crash log again, then the problem they had originally is gone surely?

C0rn3j:
> if they don't get the same crash log again, then the problem they had originally is gone surely?

They would still get a crash, just not the original one, if we don't know about a problem, we can't solve it.

And since we have no telemetry, deleting the file early could hurt us being able to know about it.

I can also imagine someone having rare crashes using some niche feature, off-hand mention it and if we then ask them for logs, we get nothing because it last happened X weeks/months ago.

I don't see any positives with deleting the file early, so I'd still say that a very long retention time since last crash is a good idea.

rexendevar:
Fair enough

PR #2201, opened 2026-06-23
fix: use original image for auto-theme color extraction

Jiafei-Queen:
## What changed

- Auto-theme sampling now uses original image instead of UI-generated thumbnail

## Fix

- Removes UI-dependent image preprocessing from auto-theme pipeline
- Keeps existing dominant-color algorithm unchanged


Taiko2k:
We don't want to do the copy and resize in the case there's no need for it. Move the `if not theme_only:​` block to after the theming block, then in the theme block, copy then at the end of it restore the `im​` reference, so subsequent code works as normal.

Issue #1338, opened 2024-12-07
Fully type the codebase


C0rn3j:
Having the codebase fully typed would prevent all kinds of issues, by being able to catch any new ones introduced by changes and any existing ones currently being hidden by lack of types.  
It also speeds up the development, especially for those unfamiliar with the codebase.

It lets us go from this:  
![image](https://github.com/user-attachments/assets/3c1d11e9-54d3-4e2e-b726-b7beaf1f578e)

To this:  
![image](https://github.com/user-attachments/assets/a52b4acf-0766-4260-8c67-5433aba646e1)

This relies on many things, and is a very tall order, but that does not mean we cannot strive to get there eventually, as we can work at this piece by piece.

### Add types for Tauon's code

This is the *easy* part, and is what I've been doing bit by bit in PRs like https://github.com/Taiko2k/Tauon/pull/1337

Pick an untyped thing, track it down as best you can and ideally type up the entire chain leading to it, and hope that it is correct, if not, it usually gets revealed later when other things get their typing.  

Typing tends to reveal bugs somewhat often in the current codebase, such as treating a variable as `list[str]` initially but later as a `str` somewhere else, which may or may not work depending on the code, but it should at the very least be cleaned up even if it technically works. Or retype such variables to expect both cases, but that usually leads to more complicated code.

### LSPs like Pyright fail to parse `t_main.py`

![image](https://github.com/user-attachments/assets/a4ca8b41-5a84-4847-9db6-9e92fea4e457)

This is partially caused by:  
 * `t_main.py` being too large - it's 49KLOC at the time of writing
 * The functions being too complex - there's a lot of conditionals that could potentially be improved

We need to split and de-complexify codebase to the point where LSPs don't time out on it - this is hinging on having types in the first place as refactoring anything without them is not a sane idea.  

For example, variable `playlist` in the codebase is sometimes an object containing among others a  `playlist: list[int]`, and sometimes it is simply an int, so moving things around without knowing what they are will be guaranteed to break things.  
This specific `playlist` example has been mostly already taken care of.

So the approach to type up the part you want to refactor (as a submodule) first is generally the go to idea.

#### Making sure we don't break things while fixing things up

It would be best if we added CI tests as we go, tracked in #1315

### Some libraries we depend on do not have (full) typing support

[...]

PR #2208, opened 2026-06-30
Fix Tidal return-type annotations: return_list yields track ids (list[int])

rudi193-cmd:
Small typing correction toward #1338.

`Tidal.fav_albums`, `fav_tracks`, `playlist`, `mix`, and `artist` are annotated `-> list[TrackClass] | None`, but in their `return_list=True` branch they build and return a list of **track ids** (`nt.index`, an `int`), not `TrackClass` objects.

The call sites confirm this — `t_main.py` does `playlist.extend(self.tidal.<m>(return_list=True))` into a list of ids, and the sibling integrations (`subsonic.get_music3`, `plex.get_albums`) already annotate the same pattern as `list[int] | None`. Tidal was the lone outlier.

Corrects the five annotations to `list[int] | None`. No runtime change (`from __future__ import annotations`); resolves 5 mypy `[return-value]` errors in this module.

### DIFF
--- a/src/tauon/t_modules/t_tidal.py
+++ b/src/tauon/t_modules/t_tidal.py
@@ -221,7 +221,7 @@ class Tidal:
 		self.tauon.gui.pl_update += 14
 		return None
 
-	def fav_albums(self, return_list: bool = False) -> list[TrackClass] | None:
+	def fav_albums(self, return_list: bool = False) -> list[int] | None:
 		self.try_load()
 		if not self.session:
 			return []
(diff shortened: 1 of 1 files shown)


rudi193-cmd:
Context for this fix: Tauon declares `Typing :: Typed` and is almost fully annotated, but there's no type-check step in CI, so drift like this slips through until someone runs mypy by hand.

Running mypy over the package, the rest of `t_tidal.py`'s annotations are otherwise consistent — the only other things it flags here are unguarded `self.session` accesses in the login/stream paths, which are currently safe because the callers guard them (a checker just can't see that), so nothing to fix there.

If you'd ever be interested in a lightweight mypy gate — even scoped to the already-typed modules — to catch annotation drift on future PRs, I'd be happy to put one together. No worries if it's not wanted.

C0rn3j:
Thanks - we already do run an LSP in the CI, but the issues are so many and not dependent on us so it's set to always pass.

The biggest offender is PySDL3 being way too strict, so effectively almost all SDL calls have a typing issue.

https://github.com/Taiko2k/Tauon/actions/runs/25058966049/job/73407272721

PR #2237, opened 2026-07-14
touch input right-click duration indicator

### LINKED ISSUE #2228: touch - visual indicator for right-click status
with my new PR I've added proper tap/rightclick/scroll/dragndrop logic. it would be nice to see visually how long you have to hold and when you've done it. My idea:

- square that displays above the touch input starting after 100 ms
- starts empty but fills up like a little progress bar
- its color changes when it's full (from button hover to button active?)
- vanishes when input switches to dragndrop or input is released

this probably can be added to the TouchInputTracker class

rexendevar:
resolves #2228


Taiko2k:
`self.gui.update += 1` is now `gui.request_frame()` FYI

PR #2242, opened 2026-07-15
2 smallish optimizations

rexendevar:
caches info for large blocks of text (eg unsynced lyrics, artist bios) & does sleeps between frames depending on the actual frame speed

### DIFF
--- a/src/tauon/t_modules/t_draw.py
+++ b/src/tauon/t_modules/t_draw.py
@@ -334,10 +334,12 @@ class TDraw:
 		cache_key = None
 		if len(text) <= 256:
 			cache_key = (text, font, max_x, wrap)
-			cached = self.text_wh_cache.get(cache_key)
-			if cached is not None:
-				self.text_wh_cache.move_to_end(cache_key)
-				return cached
+		else:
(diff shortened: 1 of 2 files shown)


src/tauon/t_modules/t_draw.py
```diff
@@ -334,10 +334,12 @@ def get_text_wh(self, text: str, font: int, max_x: int, wrap: bool = False) -> t
 		cache_key = None
 		if len(text) <= 256:
 			cache_key = (text, font, max_x, wrap)
-			cached = self.text_wh_cache.get(cache_key)
-			if cached is not None:
-				self.text_wh_cache.move_to_end(cache_key)
-				return cached
+		else:
+			cache_key = (text[:255], font, max_x, wrap)
```
Taiko2k: Actually we could drop the `[:255]` here, since text references are just pointers in Python anyway. Also just remove the whole len test block here, idk why I did that in the first place.

Taiko2k (changes requested):

Taiko2k:
Also remove the sleep change, im going to remove/refactor that myself soon.

PR #2244, opened 2026-07-16
Allow force-caching of certain large text textures

rexendevar:
increases framerate and decreases cpu usage at the cost of taking more memory. Technically this could be a big deal and we do need some kind of hard cap but for 99% of cases this is a no brainer


rexendevar:
in testing this is cutting the time it takes to run lyricsrenmini's render function from 21500 microseconds to 513 microseconds for the text render line

Taiko2k:
Ive gone and committed a fix that just removes the need for force cache

PR #2252, opened 2026-07-18
minor fix for clean venv run

rexendevar:


Taiko2k:
This works because it triggers pysdl3's binary downloader, but I don't ever want that. Really questionable security practice.

PR #2257, opened 2026-07-19
Don't raise error when SDL_CreateTray failed

City-busz:
This allows to run Tauon without AppIndicator library is installed.

### DIFF
--- a/src/tauon/t_modules/t_main.py
+++ b/src/tauon/t_modules/t_main.py
@@ -6917,7 +6917,9 @@ class Tauon:
 		# Create tray (no icon, just tooltip)
 		self.sdl_tray = sdl3.SDL_CreateTray(icon_surface, b"Tauon Music Box")
 		if not self.sdl_tray:
-			raise RuntimeError(f"SDL_CreateTray failed: {sdl3.SDL_GetError().decode()}")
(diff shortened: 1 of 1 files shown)


C0rn3j:
Ello, this will actually resolve itself even without the patch - we switched to native SDL tray implementation - but until SDL 3.6.0 is out, it uses appindicator internally anyways.

But how did you manage to run into it? Appindicator is an installed/required dep everywhere.

*Side note - Checked your [linked website](https://pkgbuild.com/~bgyorgy/) out and it still mentions the now-dead [community] instead of [extra]*

City-busz:
I removed AppIndicator library from my system to test if it can be optional. AppIndicator also pulls in the GTK3 toolkit as dependency, which is not directly used by Tauon, so it would be preferable to make it optional.

C0rn3j:
> AppIndicator also pulls in the GTK3 toolkit as dependency, which is not directly used by Tauon

It actually is - look for `gtk_settings` in `src/tauon/t_modules/t_main.py`.

Granted, it's used for very little, and I've been eyeing reimplementing it and removing it.

City-busz:
> It actually is - look for `gtk_settings` in `src/tauon/t_modules/t_main.py`.

It's in a try-except block, so the application still runs if GTK3 is not installed.

PR #2280, opened 2026-07-30
Touch: swipe sideways to middle click (queue)

### LINKED ISSUE #2239: touch: swipe sideways in playlist view to queue
Requires #2237 to be merged before i'm willing to modify TouchInputTracker again.

When in the playlist view, swiping sideways on a track or album should be equivalent to middle-clicking and should add it to the queue. We might as well do it in the SDL event processing section so it's the same input everywhere just as a fallback for other views.

- Sideways is defined as having an abs(cosine) above 0.9, so within about 26 degrees from perfectly flat
- this will evaluate in the first 10 pixels, same as scroll (unless that sucks)
-  work out hwo to move the cursor back to where it was

rexendevar:
resolves #2239. you can swipe sideways on a track in the playlist or album in the gallery to queue it, but for lack of a better input method it'll still do a middle click anywhere in the program. currently displays a fairly barebones indicator: capital letter Q in a box above the touch position, plus a small square indicating where the middle click will apply. I'm willing to rework the indicator if anyone can think of a good replacement.

update: added sideways scrolling support for the Albumflow widget and any future additions that use sideways scrolling. in these views you swipe vertically to middle click. best i could do

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 Changelog
 ---------
 
-### v11.1.2
+### v11.2.0
 
+ - Added new widget AlbumFlow
+ - Added new widget Tab Switcher
  - Revamps & refinements to both synced and static lyrics editors
  - Fixed bug that caused GNOME to occasionally crash when Tauon did
  - Touch shortcut: swipe sideways to queue (middle click)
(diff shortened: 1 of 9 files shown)


Taiko2k:
no.... no I hate it. Free flow drag should be touch only, scroll input should be by album as it was.

rexendevar:
scroll input is by album, it snaps when you release it

PR #2281, opened 2026-07-31
Tauon native entrypoint (experimental)

Taiko2k:
I've been in the AI lab cooking something HORRBILE... something CURSED!!

This PR replaces `tauonmb.sh`, and to an extent the role of `__main__.py` as the entry point. Instead we have a native binary as the entry point. So Tauon instead of being a Python app is now a C++ app that embeds Python.

Motivations:
 1. It reduces external dependencies. This PR removes `pysdl3` and `pyinstaller`.
 2. It reduces launch to splash time. (In my testing by about 200ms). This keeps IPC transfer times low for all platforms instead of using the Linux only `.sh` files.

Downside: It introduces a lot of technicality, as we are now using a custom package script in place of pyinstaller.


Solarunit:
Tauon in rust when :upside_down_face:

Taiko2k:
@Solarunit May b one day. Probably c++ would be better tho. This direction opens the path for a more staged optimisation into c++ that fits Tauons architecture.

Solarunit:
no idea if rust or c++ would be better, but SDL UI should stay! :slightly_smiling_face:

Issue #2264, opened 2026-07-24
Songs end 2 seconds early


maxgriese:
**Desktop**
- OS: Arch Linux
- OS version: Kernel 7.1.4-arch1-1
- Tauon Version: 11.1.1
- Installation method: Arch repositories/Pacman

**Describe the bug**

Tauon is connected to my jellyfin library that only consists of .flac files, but most albums are transcoded as OPUS .ogg files at 256 kbps. Playing any song regardless of jellyfin/transcode skips the last ~ 2 seconds of the song. Changing loop/ shuffle does not change this behavior. "Fade on pause and stop" and "Fade on track jump" are disabled, turning "Avoid resampling" on or off has no influence.  
This might be related to #2210 but seems like a different issue, as I do not see any weird progress bar behavior.

**To Reproduce**

1. Disable any fade options
2. Click on any song
3. Skip forward to the last ~ 10 seconds
4. Before the song finishes at ~ 2 seconds before the end, the next one is played.

**Expected behavior**

Songs play normally without skipping the end

**Screenshots/Videos**

**Tauon log**
<details><summary>Logfiles</summary>

[tauon.log](https://github.com/user-attachments/files/30352502/tauon.log)

Kevinhykuo:
Does the bug on your end only happen to tracks higher than the standard 16-Bit 44.1kHz resolution on PipeWire? For me, the bug only appears when playing high resolution tracks through Pipewire. Switching to PulseAudio fixes it.

maxgriese:
For me it happens with the standard resolution, too. Switching between pipewire and pulseaudio does not make a difference in my case.

maxgriese:
Okay, I have looked into this myself and found out that my specific problem can be worked around by enabling "Wait for entire file" in the settings. 

This seems to specifically be an issue for network tracks when there are local transcodes of the songs available. In t_phazor.py gapless handover happens if 

- a track is not a network track, or
- its url is ready, or
- it's already cached by tauon's temporary cache

This completely ignores local transcodes created by Tauon. Should be a one-line fix, I will probably create a PR for this.

PR #2339, opened 2026-09-18
Fix gapless transition when using transcodes

maxgriese:
I am encountering a bug, which leads to songs ending abruptly ~2 seconds early (Issue #2264). 

There are two effects I observed:

1. Network tracks with local transcodes end strictly 2 seconds before they should (tested using a jellyfin server).
2. Playing the transcoded tracks directly by playing them from the "Import Music Folder" playlist works and tracks play seamlessly after another. However, this still has a visual bug, where the progress bar + timer end 2 seconds earlier and the next song is already being displayed at 0:00 for ~ 2 seconds.

I noticed that in case of the network files I would get "Transition jump" for every single song transition. It seems like `target_gapless_ready` in `t_phazor.py` would never be fulfilled in this case.

I fixed this by adding the optional condition of `cachement.get_local_instant(target_object) is not None` to `target_gapless_ready`. This fixes both of the bugs I observed.


C0rn3j:
image

Wrong indents, this won't run.

maxgriese:
Should be fixed now, was a weird quirk of my setup

PR #2342, opened 2026-09-23
improve OpenSubsonic support

lc-guy:
I have 135k tracks in my navidrome server, so sync would take a _long_ time, and tauon would resync every time I edited the generator query to filter things down.

I implemented the OpenSubsonic search3 paginated endpoint which gets all the data we need much faster, and added an (optional, default off) in-memory cache which fetches the first time we use `air` in a generator after boot, and keeps it around afterwards so subsequent query edits don't constantly pull.

I'm also pulling the last modified date from subsonic and adding it to the tracks, plus added an equivalent to the `m>` / `m<` sort that allows for filtering by last modified too, like `m>2026-01-01`.

Bumped the minimum python version from 3.10 to 3.11 since a `datetime` stdlib method I needed wasn't available prior to that, and 3.10 goes EOL on October 31st anyway, but tell me if you'd rather not.


C0rn3j:
> Bumped the minimum python version from 3.10 to 3.11 since a datetime stdlib method I needed wasn't available prior to that, and 3.10 goes EOL on October 31st anyway, but tell me if you'd rather not.

Completely fine to bump to 3.11.

We're practically only held to 3.13 with the way Flatpak is currently packaged, otherwise we wouldn't really hurt by going straight to 3.14 - keep it at 3.11, though.

